PR #5, opened 2022-03-07
Move deps from lmzuccarelli to openshift

josefkarasek:


lmzuccarelli (commented):
Well spotted Josef.  As you mentioned this was reverted in PR#3. This proves to always look at the files that have changed

PR #8, opened 2022-03-09
Update makefile

lmzuccarelli:
This commit includes 

- adding the support files in the hack directory 
- updating the golangci.yaml file (disable goimports -local)
- testing the make verify
- updating the Makefile (adding verify target)


josefkarasek (approved):
lgtm

PR #12, opened 2022-03-15
Trigger profiling via http

josefkarasek:


pkg/operator/controller/nodeobservabilityrun/controller.go
```diff
@@ -18,58 +18,124 @@ package nodeobservabilityruncontroller
...
-func (r *NodeObservabilityRunReconciler) handleInProgress(instance *nodeobservabilityv1alpha1.NodeObservabilityRun) error {
-	return nil
+func (r *NodeObservabilityRunReconciler) handleInProgress(instance *nodeobservabilityv1alpha1.NodeObservabilityRun) (requeue bool, err error) {
+	var code int
+	for _, agent := range instance.Status.Agents {
+		_, code, err = httpGetCall(fmt.Sprintf("http://%s:%d/status", agent.IP, agent.Port))
+		if err != nil {
+			// TODO: node consistently failing, restart whole process
+			return
+		}
```
sherine-k, line 107: Agent also returns HTTP 500 in case the agent is "stuck" in error.

pkg/operator/controller/nodeobservabilityrun/controller.go
```diff
@@ -18,58 +18,124 @@ package nodeobservabilityruncontroller
...
+		return nil, err
+	}
+	subset := endps.Subsets[0]
+	port := subset.Ports[0].Port
+
+	targets := []nodeobservabilityv1alpha1.AgentNode{}
+	for _, a := range subset.Addresses {
+		r.Log.V(3).Info("startRun - address for loop", "IP", a.IP)
+		body, code, err := httpGetCall(fmt.Sprintf("http://%s:%d/pprof", a.IP, port))
+		if err != nil || code != 200 {
```
sherine-k, line 130: Open question: is there any gain in noting this IP down in a UnavailableTargets in the status, and eventually retrying it later within the handleInProgress?

sherine-k, line 130: Open question: is there any gain in noting this IP down in a UnavailableTargets in the status, and eventually retrying it later within the handleInProgress?

lmzuccarelli: I think it's good to have the failing ep's listed in the status. IMHO for MVP we can leave it as is, i.e not worry about re-try logic etc.

lmzuccarelli (commented):
@josefkarasek - Overall this is a neat, simple solution. Great work :)

PR #14, opened 2022-03-16
Disable  Election

sherine-k:


config/manager/controller_manager_config.yaml
```diff
@@ -7,5 +7,5 @@ metrics:
 webhook:
   port: 9443
 leaderElection:
-  leaderElect: true
+  leaderElect: false
   resourceName: 94c735b6.olm.openshift.io
```
sherine-k, line 11: Is it probably better to just remove the whole bunch?

PR #19, opened 2022-03-23
Add dockerfile for CI build

sherine-k:


Dockerfile.rhel8
```diff
@@ -0,0 +1,26 @@
+# Build the manager binary
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.17-openshift-4.10 AS builder
```
lmzuccarelli, line 2: question: about the 4.10 - as per our thread do we need to change for every release ?

Dockerfile.rhel8
```diff
@@ -0,0 +1,26 @@
...
+COPY vendor/ vendor/
+COPY go.mod go.mod
+COPY go.sum go.sum
+
+# Build
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager cmd/node-observability-operator/main.go
+
+# Use distroless as minimal base image to package the manager binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM registry.ci.openshift.org/ocp/4.10:base
```
lmzuccarelli, line 21: same here

Dockerfile.rhel8
sherine-k: From  the info I've got , yes : every OCP release the dependencies and the Dockerfile should be updated.

PR #20, opened 2022-03-24
Add retry logic

josefkarasek:


lmzuccarelli (commented):
small fixes if you don't mind 
1. Define a constant instead of duplicating this literal 3 times i.e "Instance"	r.Log.V(3).Info("Reconcile", "Instance", req.NamespacedName.String())
2. If possible do it for other repeats as well 

For code smell explanation 
```
Duplicated string literals make the process of refactoring error-prone, since you must be sure to update all occurrences.
On the other hand, constants can be referenced from many places, but only need to be updated in a single place.
```

pkg/operator/controller/nodeobservabilityrun/error.go
```diff
@@ -0,0 +1,23 @@
...
+func (e NodeObservabilityRunError) Error() string {
+	return fmt.Sprintf("Code: %d, Error: %s", e.HttpCode, e.Msg)
+}
+
+func IsNodeObservabilityRunErrorRetriable(err error) bool {
+	e, ok := err.(NodeObservabilityRunError)
+	if !ok {
+		return false
+	}
+	if e.HttpCode == 500 {
```
josefkarasek, line 19: use constant from net/http

pkg/operator/controller/nodeobservabilityrun/controller.go
```diff
@@ -165,23 +168,28 @@ func inProgress(instance *nodeobservabilityv1alpha1.NodeObservabilityRun) bool {
...
+		defer resp.Body.Close()
 
-	body, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
-		return "", 0, err
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return err
+		}
+		if resp.StatusCode != 200 {
```
josefkarasek, line 188: use constant

pkg/operator/controller/nodeobservabilityrun/error.go
sherine-k: when the agent returns 500, in most cases, with the current implementation, it is "stuck" until an admin cleans up the error file.
I don't know if we want to keep retrying in this case.
Do we want to look at httpCode 409? It's returned when the agent is busy by another profiling request, and will be able to run this request a bit later.

PR #29, opened 2022-04-06
Activate gosec in golangci

sherine-k:
* Annotate constant SecretName in vars.go as a false positive
* Fix the false positive that was also highlighted during the audit review


lmzuccarelli:
@sherine-k - this is extremely valuable. Thanks for doing this

PR #38, opened 2022-04-25
Fix bundle generation

thejasn:
Fixes and cleans up manifest generation
---
Changes 

- `Makefile` : Add new recipe `operator-sdk` as a prerequisite to `make bundle` which downloads `operator-sdk` to *./bin/* directory and pins the SDK version to v1.19.0. 
- `bundle/manifests` : Clean up of old generated resources (mostly from `make olm-manifests`).
- `manager.yaml` : Add namespace creation during CRD creation. 
- `manager.yaml`and `kustomization.yaml` : Fixed image replacement done using `$(KUSTOMIZE) edit set image controller=$(IMG)` by changing placeholder image `registry.ci.openshift.org/ocp/4.11:node-observability-operator` to `quay.io/openshift/node-observability-operator:v0.0.1` to make sure it gets replaced in the CI according to [config here](https://github.com/openshift/release/blob/72754375ef5ae77d220094a9b8b0b8091af35687/ci-operator/config/openshift/node-observability-operator/openshift-node-observability-operator-main.yaml#L37).

Signed-off-by: thejasn <thn@redhat.com>


sherine-k (approved):
Thanks @thejasn !

This really tidies up the makefile and manifests, and solves the issue with the ci image which wasn't working locally for me.

PR #43, opened 2022-05-09
Adding unit test for NOR controller

josefkarasek:


sherine-k (commented):
Great job @josefkarasek 
Are you expecting to be able to test the MCO related code by E2E instead?
We could be missing some tests on HTTP error responses from the agents though. What do you think?

PR #45, opened 2022-05-11
Clean up resources created and restore cluster state

bharath-b-rh:
PR:
- When profiling config is unset, resources created for the same will be cleaned up after MCO's fallback to previous rendered MC is successful
- Label worker nodes with custom label to use in node selector.
- Enable Config flow updated
  Label nodes
  Create MCP
  Create MC
- Disable Config flow updated
   Unlabel nodes
   Wait for worker MCP to be updated
   Remove MC and MCP created.
   
 **Unit test scenarios are commented as majority code has been updated.**


pkg/operator/controller/machineconfig/constants.go
```diff
@@ -0,0 +1,110 @@
...
+
+const (
+	// finalizer name for nodeobservabilitymachineconfig resources
+	finalizer = "NodeObservabilityMachineConfig"
+
+	// defaultRequeueTime is the default reconcile requeue time
+	defaultRequeueTime = 3 * time.Minute
+
+	// Null is defined for empty string
+	Null = ""
```
josefkarasek, line 31: that's confusing
it's either nil or "" in go

bharath-b-rh: changed from Null to Empty.

pkg/operator/controller/machineconfig/controller_test.go
```diff
@@ -16,11 +16,11 @@ limitations under the License.
 
 package machineconfigcontroller
 
+/*
```
saschagrunert, line 19: Do we wanna remove this one?

pkg/operator/controller/machineconfig/crio.go
```diff
@@ -32,84 +33,77 @@ import (
...
 
-// ensureCrioProfConfigExists checks if CRI-O MachineConfig CR for
-// enabling profiling exists, if not creates the resource
-func (r *MachineConfigReconciler) ensureCrioProfConfigExists(ctx context.Context) (*mcv1.MachineConfig, bool, error) {
-	namespace := types.NamespacedName{Name: CrioProfilingConfigName}
-	criomc, exists, err := r.fetchCrioProfileConfig(ctx, namespace)
+	criomc, err := r.fetchCrioProfConf(ctx, namespace)
 	if err != nil {
-		return nil, false, err
+		return createdNow, fmt.Errorf("failed to ensure crio profiling config was indeed created: %v", err)
```
saschagrunert, line 55: nit, let's wrap those errors:
```suggestion
		return createdNow, fmt.Errorf("failed to ensure crio profiling config was indeed created: %w", err)
```

pkg/operator/controller/machineconfig/mcp_test.go
```diff
@@ -16,6 +16,7 @@ limitations under the License.
 
 package machineconfigcontroller
 
+/*
```
saschagrunert, line 19: What about those?

saschagrunert (commented):
Great work @bharath-b-rh! I'm keen to iterate and merge this one sooner than later. I just have a question about the unit tests.

pkg/operator/controller/machineconfig/controller_test.go
bharath-b-rh: Since there are new changes added, thought to make use of unit test story to add new cases and existing cases also require changes, so commented those, to rework everything at once.

pkg/operator/controller/machineconfig/mcp_test.go
bharath-b-rh: Since there are new changes added, thought to make use of unit test story to add new cases and existing cases also require changes, so commented those, to rework everything at once.

saschagrunert (approved):

lmzuccarelli:
@bharath-b-rh - agree with @saschagrunert - great work.

lmzuccarelli:
Had a look at the logs for the e2e test seem like its an internal gcp error

pkg/operator/controller/machineconfig/mcp.go
```diff
@@ -106,91 +122,204 @@ func (r *MachineConfigReconciler) GetProfilingMCP() *mcv1.MachineConfigPool {
...
+// createMCP is for creating the required MCP
+func (r *MachineConfigReconciler) createMCP(ctx context.Context, name string) error {
+	mcp := r.GetProfilingMCP(name)
 
 	if err := r.Create(ctx, mcp); err != nil {
-		return fmt.Errorf("failed to create MCP for profiling machine configs: %w", err)
+		return fmt.Errorf("failed to create MCP %s: %w", name, err)
 	}
 
 	if err := ctrl.SetControllerReference(r.CtrlConfig, mcp, r.Scheme); err != nil {
```
josefkarasek, line 139: This looks like a noop to me.

The resource is modified in local memory, but it's not persisted. Move it just a few lines above.

pkg/operator/controller/machineconfig/crio.go
```diff
@@ -32,84 +33,77 @@ import (
...
-		if err != nil || exists {
-			return false, fmt.Errorf("failed to ensure crio profiling config was indeed deleted: %v", err)
-		}
+	if err := r.deleteCrioProfConf(ctx, criomc); err != nil {
+		return err
+	}
 
-		return true, nil
+	r.Lock()
+	r.MachineConfig.PrevReconcileUpd["crio"] = MachineConfigInfo{
```
josefkarasek, line 87: This panic is possible here:
```
panic: assignment to entry in nil map

goroutine 1 [running]:

Program exited.
```
Please:
* Mitigate the potential of this panic
* use defer for r.Unlock() (defer is called even during a panic)

pkg/operator/controller/machineconfig/crio.go
```diff
@@ -32,84 +33,77 @@ import (
...
-		}
-
-		return criomc, true, nil
+	r.Lock()
+	r.MachineConfig.PrevReconcileUpd["crio"] = MachineConfigInfo{
+		op:     "create",
+		config: criomc,
 	}
-	return criomc, false, nil
+	r.Unlock()
```
josefkarasek, line 63: Same `panic` issue here

pkg/operator/controller/machineconfig/crio.go
```diff
@@ -32,84 +33,77 @@ import (
...
+		createdNow = true
+	}
 
-	// CrioUnixSocketConfData contains the configuration required
-	// for enabling CRI-O profiling
-	CrioUnixSocketConfData = `[Service]
-Environment="ENABLE_PROFILE_UNIX_SOCKET=true"`
-)
+	// grace time for client cache to get refreshed
+	time.Sleep(500 * time.Millisecond)
```
josefkarasek, line 51: Why do you think this is needed?

pkg/operator/controller/machineconfig/crio.go
bharath-b-rh: Map is initialized [here](https://github.com/openshift/node-observability-operator/pull/45/files/aeb062be957c34efd81f81379613c81b00f2148a#diff-00a14626d7a56d8be3fe940ee9c4ffbdf0676de1f64453f29ae987c4f16961e3R149), when reconciler object is created and is not freed in the code, nil map panic should not be observed. But using defer is better approach, have incorporated comment.

pkg/operator/controller/machineconfig/crio.go
bharath-b-rh: comment incorporated, used defer for releasing lock.

pkg/operator/controller/machineconfig/crio.go
bharath-b-rh: At certain times observed client cache not getting updated and immediate fetch was failing, adding delay here solved that.

lmzuccarelli:
@bharath-b-rh - finally all tests passing

pkg/operator/controller/machineconfig/controller.go
```diff
@@ -19,96 +19,29 @@ package machineconfigcontroller
...
-//+kubebuilder:rbac:groups=machineconfiguration.openshift.io,resources=machineconfigpools,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=machineconfiguration.openshift.io,resources=machineconfigs,verbs=get;list;create;delete
+//+kubebuilder:rbac:groups=machineconfiguration.openshift.io,resources=machineconfigpools,verbs=get;list;create;delete
+//+kubebuilder:rbac:groups=core,resources=nodes,verbs=get;list;patch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
 //
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.0/pkg/reconcile
```
josefkarasek, line 43: this comment is auto-generated, pls add a more specific doc text to this controller

pkg/operator/controller/machineconfig/crio.go
josefkarasek: I think you have a fairly up-to-date obj already:

https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/typed_client.go#L51

pkg/operator/controller/machineconfig/mcp.go
bharath-b-rh: comment incorporated

pkg/operator/controller/machineconfig/controller.go
bharath-b-rh: comment incorporated

PR #47, opened 2022-05-16
Update README.md

bharath-b-rh:
Removed $ from each of the command, to copy and use commands as is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,23 +15,23 @@ You can install the NodeObservability Operator by building and pushing the Opera
 1. To build and push the Operator image into a registry, run the following commands:
    ```sh
    # set the envar CONTAINER_ENGINE to the preffered container manager tool (default is podman)
-   $ export IMG=<registry>/<username>/node-observability-operator:latest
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):

darshan-nagaraj:
@bharath-b-rh It's a standard convention in Openshift docs to mention about normal user($) and root user (#). We don't mention in the prerequisites section that user should be either "$" or "#". 

Also, I would like to mention that if you use **copy to clipboard** icon from docs.openshift.com, it will copy only command and excludes "$" or "#".

bharath-b-rh:
@darshan-nagaraj Few of the openshift repositories did not have the commands with the prompt symbol, but if majority doesn't follow this convention, then better to have it standard way.
I will remove the changes and just have [this](https://github.com/openshift/node-observability-operator/pull/47/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R58) correcting the file name in command.

PR #48, opened 2022-05-19
Integration of NOR and NOMC controllers to enable/disable profiling

bharath-b-rh:
PR contains:
- NOR controller creates NOMC CR with CRIO profiling enabled.
- NOR controller monitors NOMC status to check if enabling/disabling of profiling was successful.
- NOR controller disables profiling in NOMC and checks the status. If disabling was successful, removes the created NOMC CR.


pkg/operator/controller/nodeobservabilityrun/mco.go
```diff
@@ -90,3 +90,98 @@ func (r *NodeObservabilityRunReconciler) desiredMCO() *v1alpha1.NodeObservabilit
...
+func (r *NodeObservabilityRunReconciler) createNOMC(ctx context.Context, mc *v1alpha1.NodeObservabilityMachineConfig) error {
+	if err := r.Create(ctx, mc); err != nil {
+		return fmt.Errorf("failed to create NodeObservabilityMachineConfig %s: %w", mc.Name, err)
+	}
+	r.Log.Info("created NodeObservabilityMachinConfig", "name", mc.Name)
+	return nil
+}
+
+func (r *NodeObservabilityRunReconciler) updateNOMC(ctx context.Context, mc *v1alpha1.NodeObservabilityMachineConfig) error {
+	if err := r.Update(ctx, mc); err != nil {
```
josefkarasek, line 130: Pls check for conflicts

pkg/operator/controller/nodeobservabilityrun/controller.go
```diff
@@ -87,23 +87,34 @@ func (r *NodeObservabilityRunReconciler) Reconcile(ctx context.Context, req ctrl
...
+		errUpdate := r.Status().Update(ctx, instance, &client.UpdateOptions{})
+		if errUpdate != nil {
+			r.Log.Error(err, "failed to update status")
+			err = utilerrors.NewAggregate([]error{err, errUpdate})
+		}
+	}()
+
 	// interact with MCO only if we are executing CrioKubeletProfile runs
 	// also bypass if the runtype is e2e-test
 	if instance.Spec.RunType != nodeobservabilityv1alpha1.E2ETest && (instance.Spec.RunType == nodeobservabilityv1alpha1.CrioKubeletProfile) {
```
josefkarasek, line 100: I'm strongly against volkswagen style testing. System Under Test should be identical to the system that will run in production env.

pkg/operator/controller/nodeobservabilityrun/controller.go
```diff
@@ -87,23 +87,34 @@ func (r *NodeObservabilityRunReconciler) Reconcile(ctx context.Context, req ctrl
...
-		result, err := r.MCOReconciler.CheckNodeObservabilityMCPStatus(ctx)
+		// this will ensure we dont trigger the agent call
+		// until the mco and mcp have been updated successfully
+
+		enabled, err := r.checkNOMCStatus(ctx, nodeobservabilityv1alpha1.DebugEnabled)
 		if err != nil {
-			r.Log.Info("waiting for NodeObservabilityMachineConfig update in NodeobeservabilityRun")
-			return result, err
+			r.Log.Error(err, "NodeObservabilityMachineConfig enable status check failed")
+			return ctrl.Result{RequeueAfter: 1 * time.Minute}, err
```
josefkarasek, line 110: Why do you want to requeue in 1 minute?

josefkarasek, line 110: Why do you want to requeue in 1 minute?

pkg/operator/controller/nodeobservabilityrun/controller.go
```diff
@@ -115,29 +126,34 @@ func (r *NodeObservabilityRunReconciler) Reconcile(ctx context.Context, req ctrl
...
+			if disabled {
+				r.Log.Info("NodeObservabilityMachineConfig CrioKubeletProfile disabled")
+				nomc := r.desiredMCO()
+				if err := r.deleteNOMC(ctx, nomc); err != nil {
+					if !errors.IsNotFound(err) {
+						return ctrl.Result{}, err
+					}
+				}
+
+				instance.Spec.RunType = ""
```
josefkarasek, line 148: This looks like a noop

josefkarasek, line 148: This looks like a noop

pkg/operator/controller/nodeobservabilityrun/controller.go
bharath-b-rh: If node labeling, MachineConfig or Pool creation fails then the same is update NOMC status, but if it was successful and MCO is updating the configs it will take more than a minute for it to finish. So thought to reconcile back again after a minute was apt.

pkg/operator/controller/nodeobservabilityrun/controller.go
bharath-b-rh: This needs to be discussed, shouldn't have included in PR. But thought to discuss and update it. Need a way to disable it, as during reconcile after 30minutes and removing all configs, if set everything will be repeated.

pkg/operator/controller/nodeobservabilityrun/mco.go
bharath-b-rh: comment incorporated.

pkg/operator/controller/nodeobservabilityrun/mco.go
```diff
@@ -90,3 +90,98 @@ func (r *NodeObservabilityRunReconciler) desiredMCO() *v1alpha1.NodeObservabilit
...
+	if err != nil {
+		return false, err
+	}
+
+	if len(nomc.Status.Conditions) == 0 {
+		r.Log.Info("NodeObservabilityMachineConfig has not updated any conditions yet")
+		return false, nil
+	}
+
+	for _, cond := range nomc.Status.Conditions {
```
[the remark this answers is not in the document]
bharath-b-rh: comment incorporated

pkg/operator/controller/machineconfig/mcp.go
```diff
@@ -172,6 +172,8 @@ func (r *MachineConfigReconciler) CheckNodeObservabilityMCPStatus(ctx context.Co
 		r.Log.Info("NodeObservabilityMachineConfig current condition "+
 			"does not require MCP status check",
 			"MCP", mcp.Name, "conditions", r.CtrlConfig.Status.Conditions)
+
+		r.Unlock()
```
bharath-b-rh, line 176: Current branch was updated with main, which had #49 changes, hence is listed here too.

api/v1alpha1/nodeobservabilitymachineconfig_helpers.go
```diff
@@ -119,3 +121,23 @@ func IsNodeObservabilityMachineConfigConditionPresentAndEqual(conditions []NodeO
...
+				return true, nil
+			}
+		} else {
+			if cond.Status == ConditionTrue ||
+				cond.Status == ConditionInProgress {
+				return false, fmt.Errorf("NodeObservabilityMachineConfig not in expected state. Condition: %s State: %s", cond.Type, cond.Status)
+			}
+		}
+	}
+	return false, fmt.Errorf("NodeObservabilityMachineConfig in unknown state")
```
saschagrunert, line 142: nit:
```suggestion
	return false, errors.New("NodeObservabilityMachineConfig in unknown state")
```

pkg/operator/controller/nodeobservabilityrun/controller.go
bharath-b-rh: comment incorporated

lmzuccarelli:
> Hey @saschagrunert, @bharath-b-rh: I wouldn't tie the MachineConfig life-cycle the the NodeObservabilityRun. NodeObservabilityRuns are supposed to be cheap operations, that you can run every 30 seconds or so.
> 
> NodeObservability is a more longetive resource, that encapsulates expensive operations. Right now, it manages the life cycle of node agents. Adding node configuration here would make more sense to me.

@josefkarasek - the discussion we had with @saschagrunert regarding this is that other type of runs i.e eBPF and bash will more than likely not need MCO, so thats why we added it in the NOR. It makes no sense to enable MCO if don't need it. Lets discuss this and come up with a joint solution

api/v1alpha1/nodeobservabilitymachineconfig_helpers.go
```diff
@@ -119,3 +121,23 @@ func IsNodeObservabilityMachineConfigConditionPresentAndEqual(conditions []NodeO
 	}
 	return false
 }
+
+// IsNodeObservabilityMachineConfigConditionAsExpected returns true when conditionType is present and set to `ConditionTrue`, false when not present or set to `ConditionInProgress`. Returns error when anyother conditionType is set to `ConditionTrue` or `ConditionInProgress`
+func IsNodeObservabilityMachineConfigConditionAsExpected(conditions []NodeObservabilityMachineConfigCondition, conditionType NodeObservabilityMachineConfigConditionType) (bool, error) {
+	for _, cond := range conditions {
+		if cond.Type == conditionType {
```
sherine-k, line 128: question: do we expect the matching condition to always be the first one?
Should we keep the result of the else L135 in a var till  after the loop in order to return false once all conditions were cheched?

api/v1alpha1/nodeobservabilityrun_types.go
```diff
@@ -54,6 +54,7 @@ const (
 	CrioKubeletProfile NodeObservabilityRunType = "crioKubeletProfile"
 	EbpfProfile        NodeObservabilityRunType = "eBPFprofile"
 	E2ETest            NodeObservabilityRunType = "e2e-test"
+	Disabled           NodeObservabilityRunType = "disabled"
```
sherine-k, line 57: I'm wondering if we could use another word . 
a NodeObservabilityRun of type `disabled` might be confusing to the end user.
Propositions: plain, standard, default

pkg/operator/controller/machineconfig/controller.go
```diff
@@ -204,50 +220,54 @@ func (r *MachineConfigReconciler) inspectProfilingMCReq(ctx context.Context) err
...
-		return r.revertNodeLabeling(ctx)
+		return true, r.revertNodeLabeling(ctx)
 	}
 	setEnabledCondition += modCount
 	if modCount, err = r.ensureReqMCPExists(ctx); err != nil {
 		r.Log.Error(err, "failed to ensure mcp exists")
-		return
+		return false, err
 	}
 	setEnabledCondition += modCount
```
sherine-k, line 237: Sorry, question: 
I'm having trouble understanding why we  sum the number of labeled nodes with a +1 of MCP applied, MC applied.
If the MCP or MC are not applied we'd probably get an error and exit.
But if the number of labeled nodes isn't correct, wouldn't it go unnoticed? since we're just checking if `setEnabledCondition` > 0

api/v1alpha1/nodeobservabilityrun_types.go
bharath-b-rh: Included Disabled to avoid creating NodeObservabilityMachineConfig during each reconciling once after 30s window of profiling was completed, but RestoreMCOStateAfterRun serves way for retaining the MC related changes, hence removed it.

pkg/operator/controller/machineconfig/controller.go
bharath-b-rh: Used int type instead of bool to be aware of any new changes made and need to defer the MCP check to next reconciling call. If labeling a node fails, error is returned and it's retried in the next reconciliation.

PR #50, opened 2022-05-19
Add missing mutext unlock

bharath-b-rh:


arjunrn:
Ideally unlocking is done through a [defer](https://go.dev/doc/effective_go#defer) so that even if there is a error the lock is always released before returning. And it's also more succinct to do it that way.

bharath-b-rh:
> Ideally unlocking is done through a [defer](https://go.dev/doc/effective_go#defer) so that even if there is a error the lock is always released before returning. And it's also more succinct to do it that way.

Yeah I agree, but in this scenario, method called is using lock and had to release the lock at each instance of return in this method.

PR #52, opened 2022-05-23
Improvements in NOMC controller

bharath-b-rh:
- Add RBAC for creating events to fix error `events is forbidden: User "system:serviceaccount:node-observability-operator:node-observability-operator-controller-manager" cannot create resource "events" in API group "" in the namespace "default"' (will not retry!)`, observed in cluster created using CRC.
- Fix immediate invoking of Reconcile method, which was causing worker MCP to be in degraded state. As the worker MCP would be in Updated state, immediate check after removing node labels, gave the impression of update finished, and NOMC was removing the MC and MCP created, which was causing rendered MC of nodeobservability to be removed and affecting the state of worker MCP.
- When NOMC resource has DeletionTimestamp set, as part of cleanup disable the profiling if enabled and remove finalizer once it's successful.
- Retry updating status, observed a few times it failing due to resource version change.

### DIFF
--- a/pkg/operator/controller/machineconfig/controller.go
+++ b/pkg/operator/controller/machineconfig/controller.go
@@ -372,10 +372,7 @@ func (r *MachineConfigReconciler) ensureReqMCExists(ctx context.Context) (int, e
 // ensureReqMCNotExists is for ensuring the machine config created when
 // profiling was enabled is indeed removed
 func (r *MachineConfigReconciler) ensureReqMCNotExists(ctx context.Context) error {
-	if !r.CtrlConfig.Spec.Debug.EnableCrioProfiling {
-		return r.disableCrioProf(ctx)
-	}
-	return nil
+	return r.disableCrioProf(ctx)
 }
 
 // ensureReqMCPExists is for ensuring the required machine config pool exists
--- a/pkg/operator/controller/machineconfig/mcp.go
+++ b/pkg/operator/controller/machineconfig/mcp.go
@@ -265,7 +265,7 @@ func (r *MachineConfigReconciler) checkWorkerMCPStatus(ctx context.Context) (ctr
 	if mcv1.IsMachineConfigPoolConditionTrue(mcp.Status.Conditions, mcv1.MachineConfigPoolUpdated) {
 
 		r.Unlock()
-		if err := r.disableCrioProf(ctx); err != nil {
+		if err := r.ensureReqMCNotExists(ctx); err != nil {
 			return ctrl.Result{RequeueAfter: 1 * time.Minute}, err
 		}
 		if err := r.ensureReqMCPNotExists(ctx); err != nil {


api/v1alpha1/nodeobservabilitymachineconfig_helpers.go
```diff
@@ -20,6 +20,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// UpdateLastReconcileTime is for updating LastReconcile in NodeObservabilityMachineConfigStatus
+func UpdateLastReconcileTime(status *NodeObservabilityMachineConfigStatus) {
```
saschagrunert, line 24: How about binding the function to become a method of `NodeObservabilityMachineConfigStatus`?
```suggestion
func (status *NodeObservabilityMachineConfigStatus) UpdateLastReconcileTime() {
```

pkg/operator/controller/machineconfig/controller.go
```diff
@@ -36,6 +37,7 @@ import (
 //+kubebuilder:rbac:groups=machineconfiguration.openshift.io,resources=machineconfigs,verbs=get;list;watch;create;delete
 //+kubebuilder:rbac:groups=machineconfiguration.openshift.io,resources=machineconfigpools,verbs=get;list;watch;create;delete
 //+kubebuilder:rbac:groups=core,resources=nodes,verbs=get;list;watch;patch
+//+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
```
saschagrunert, line 40: For what do we need to patch events?

saschagrunert (commented):
Just two nits, otherwise LGTM :)

api/v1alpha1/nodeobservabilitymachineconfig_helpers.go
josefkarasek: We already have a timestamp associated with the InProgress condition:
https://github.com/openshift/node-observability-operator/pull/52/files#diff-414059186bd44b4fc1458ec60a47e34426313d2470846d8e8d71a48d7f7506d3R90

Can that be used?

pkg/operator/controller/machineconfig/controller.go
bharath-b-rh: We do not need permission for Patch on Events, will correct it.

api/v1alpha1/nodeobservabilitymachineconfig_helpers.go
bharath-b-rh: Yeah, but I thought it will be helpful to have a separate parameter in case we need it for debugging, since nothing is logged if the cluster is in desired state.

bharath-b-rh: comment incorporated

pkg/operator/controller/machineconfig/controller.go
bharath-b-rh: comment incorporated

saschagrunert (approved):
/lgtm

PR #55, opened 2022-05-25
Fix image used in e2e

sherine-k:


josefkarasek:
@lmzuccarelli any experience with some linter/check that we don't accidentally override images?

PR #57, opened 2022-05-26
Fixed FIXMEs

bharath-b-rh:


pkg/operator/controller/machineconfig/controller.go
```diff
@@ -343,11 +344,11 @@ func (r *MachineConfigReconciler) ensureProfConfDisabled(ctx context.Context) (b
 		return true, r.revertNodeUnlabeling(ctx)
 	}
 
-	// FIXME: shouldn't these conditions be always set to False, when ensureProfConfDisabled() is called?
```
bharath-b-rh, line 346: Yes, but DebugReady should be set to InProgress when there is a change and requires update.

josefkarasek (commented):
/approve
/lgtm
/label docs-approved
/label qe-approved
/label px-approved

PR #56, opened 2022-05-25
Add types to NodeObservability

josefkarasek:
Based on the types the node-observability-controller can provision NodeObservabilityMachineConfig, if requested type requires a change of node configuration

PR #65, opened 2022-06-07
Refactor logging: make structured, remove duplicates, formatting

alebedev87:
Follow-up of the review comments about the logging from https://github.com/openshift/node-observability-operator/pull/56.

### DIFF
--- a/pkg/operator/controller/machineconfig/controller.go
+++ b/pkg/operator/controller/machineconfig/controller.go
@@ -90,12 +90,11 @@ func (r *MachineConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
-			r.Log.Info("NodeObservabilityMachineConfig resource not found. Ignoring could have been deleted", "name", req.NamespacedName.Name)
+			r.Log.Info("NodeObservabilityMachineConfig resource not found. Ignoring could have been deleted", "Name", req.NamespacedName.Name)
 			return ctrl.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
-		r.Log.Error(err, "failed to fetch NodeObservabilityMachineConfig")
-		return ctrl.Result{RequeueAfter: defaultRequeueTime}, err
+		return ctrl.Result{RequeueAfter: defaultRequeueTime}, fmt.Errorf("failed to fetch NodeObservabilityMachineConfig: %w", err)
 	}
 	r.Log.V(3).Info("NodeObservabilityMachineConfig resource found")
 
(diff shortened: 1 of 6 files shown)


pkg/operator/controller/machineconfig/controller.go
```diff
@@ -265,7 +259,6 @@ func (r *MachineConfigReconciler) removeFinalizer(ctx context.Context, req ctrl.
 
 	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		if err := r.ClientGet(ctx, req.NamespacedName, withoutFinalizers); err != nil {
-			r.Log.Error(err, "failed to fetch nodeobservabilitymachineconfig resource for removing finalizer")
```
sherine-k, line 268: Maybe we could add the following for context: 
```
err = utilerrors.NewAggregate([]error{err, fmt.Errorf("failed to fetch nodeobservabilitymachineconfig resource for removing finalize: %w", err)}
```

pkg/operator/controller/machineconfig/controller.go
```diff
@@ -289,7 +282,6 @@ func (r *MachineConfigReconciler) removeFinalizer(ctx context.Context, req ctrl.
 
 		withoutFinalizers.Finalizers = newFinalizers
 		if err := r.ClientUpdate(ctx, withoutFinalizers); err != nil {
-			r.Log.Error(err, "failed to remove nodeobservabilitymachineconfig resource finalizers")
```
sherine-k, line 292: Maybe we could add the following for context:
```
err = utilerrors.NewAggregate([]error{err, fmt.Errorf("failed to remove nodeobservabilitymachineconfig resource finalizers: %w", err)}

File: pkg/operator/controller/machineconfig/mcp.go [hunk 3]
```diff
@@ -133,15 +133,15 @@ func (r *MachineConfigReconciler) GetProfilingMCP(name string) *mcv1.MachineConf
 func (r *MachineConfigReconciler) createMCP(ctx context.Context, name string) error {
 	mcp := r.GetProfilingMCP(name)
 
-	if err := ctrlutil.SetControllerReference(r.CtrlConfig, mcp, r.Scheme); err != nil {
-		r.Log.Error(err, "failed to update owner info in MCP", "MCP", name)
+	if err := ctrlutil.SetOwnerReference(r.CtrlConfig, mcp, r.Scheme); err != nil {
```
sherine-k, line 136: I'm having a doubt about this one. I think this was changed to SetControllerReference very recently

sherine-k (approved):
Thanks so much @alebedev87 for taking the time to make this right. 
There's just one doubt from me about SetOwnerReference instead of SetControllerReference, and we're good to go

pkg/operator/controller/machineconfig/mcp.go
alebedev87: Yes, my fault, bad merge conflict resolution. `ControllerRef` must be kept to make the cascade deletion and owner re-queuing working.

pkg/operator/controller/machineconfig/controller.go
alebedev87: This message doesn't really bring any new information: 
- if fetching fails it'll be in the error
- the fact that it's done for the removal of the finalizers will be mentioned up in the call stack (`cleanUp` method)

pkg/operator/controller/machineconfig/controller.go
alebedev87: This message contains the duplicate information, the error returned by this method will be wrapped in the error with the same message up the call stack (in `cleanUp` method).

thejasn:
For logging across multiple controllers, we should probably prefix with the controller name for better readability. 
```go
const (
	controllerName = "machineconfig_controller"
)

var log = logf.Logger.WithName(controllerName)

```
And we use this `log` instead of fetching it from context. WDYT?

sherine-k (approved):

PR #67, opened 2022-06-10
Consume installation namespace

alebedev87:
Follow-up of [this comment](https://github.com/openshift/node-observability-operator/pull/56#discussion_r887698835).


josefkarasek:
We should get ready for having `openshift-` prefix in the project name.

/cc @sherine-k

sherine-k (approved):
/lgtm

PR #72, opened 2022-06-21
nodeobservability: use controlled ownerref for daemonset

alebedev87:
DaemonSet's owner reference now has the controller set to `true` which allows to trigger the reconciliation. This helps with the cases when the daemonset was removed.


sherine-k (approved):
/lgtm

PR #75, opened 2022-06-24
rbac: move local roles into a separate file

alebedev87:


lmzuccarelli:
@alebedev87 - I would assume this is for easier maintenance and also keeps a good separation of concerns ?

lmzuccarelli:
@alebedev87 - thanks for the explanation. Yes agreed the better route is to use ` make bundle`

PR #79, opened 2022-07-05
Add local_manager_role.yaml and local_role_binding.yaml to kustomize

sherine-k:
Since https://github.com/openshift/node-observability-operator/pull/75, the local role and its bindings are not deployed when running `make deploy`.
Adding them to the config/rbac/kustomization.yaml

### DIFF
--- a/bundle/manifests/node-observability-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/node-observability-operator.clusterserviceversion.yaml
@@ -434,6 +434,76 @@ spec:
           verbs:
           - create
           - patch
+        - apiGroups:
+          - apps
+          resources:
+          - daemonsets
+          verbs:
+          - create
+          - get
+          - list
+          - watch
+        - apiGroups:
+          - ""
+          resources:
+          - configmaps
+          verbs:
+          - create
+          - delete
+          - get
+          - list
+          - update
+          - watch
+        - apiGroups:
+          - ""
+          resources:
+          - secrets
+          verbs:
+          - create
+          - delete
+          - get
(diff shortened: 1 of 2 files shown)


lmzuccarelli:
@sherine-k - seems to be a verify problem on `make bundle`  - could you have a look and update the PR please

sherine-k:
> @sherine-k - seems to be a verify problem on `make bundle` - could you have a look and update the PR please

Done. Can you check with me on the CSV please?
It added a chunk of things... under permissions. 
Which is normal I guess, except for the duplication of rbac regarding configmaps.
I'm not sure where that is coming from.
@sunilcio any rbac issues you might have noticed on the last bundle install from CVP?

bharath-b-rh (approved):
lgtm

bundle/manifests/node-observability-operator.clusterserviceversion.yaml
```diff
@@ -434,6 +434,76 @@ spec:
           verbs:
           - create
           - patch
+        - apiGroups:
```
alebedev87, line 437: We don't need these files anymore then:

- https://github.com/openshift/node-observability-operator/blob/bf1a0755825a1da11dc2a50238f024c23eb098b9/bundle/manifests/node-observability-operator_rbac.authorization.k8s.io_v1_role.yaml
- https://github.com/openshift/node-observability-operator/blob/bf1a0755825a1da11dc2a50238f024c23eb098b9/bundle/manifests/node-observability-operator_rbac.authorization.k8s.io_v1_rolebinding.yaml

PR #83, opened 2022-08-09
CFE-500

ayesha54:
Details on the implementation:

- NodeObservability API has been updated to have .spec.nodeSelector field instead of .spec.labels to reflect the meaning better behind
- NodeObservabilityMachineConfig API has been updated to inherit nodeSelector from NodeObservability CR
- NodeObservability's node selector has been used in the agent's DaemonSet.
- UT Commented, UT scenario written for patching nodes with the custom label is failing, which will be not required, the existing worker label will be used.


pkg/operator/controller/machineconfig/nodes.go
```diff
@@ -93,27 +93,28 @@ func (r *MachineConfigReconciler) ensureReqNodeLabelExists(ctx context.Context)
...
-		}
-		r.Log.V(1).Info("Successfully added label", "Node", node.Name, "Label", NodeObservabilityNodeRoleLabelName)
+			patch, _ := newPatch(add,
+				ResourceLabelsPath,
+				map[string]interface{}{
+					nodeSelectorKey: nodeSelectorValue,
+				})
+			if err := r.ClientPatch(ctx, &nodeList.Items[i], client.RawPatch(types.JSONPatchType, patch)); err != nil {
+				return updNodeCount, err
+			}
```
sherine-k, line 109: Hi @ayesha54 
I'm not sure if we should add the nodeSelector to all worker nodes: the admin must have already set the right selectors on those nodes he/she likes to perform the pprof on. 
So, instead, I think we should  use it in order to identify nodes to include in the machineConfigPool.
WDYT?

sherine-k (changes requested):
Hi @ayesha54 
This is a great draft for CFE-500. 
Keep it up!

pkg/operator/controller/machineconfig/nodes.go
bharath-b-rh: Hi @ayesha54 
Methods used for ensuring and removing the labels will not be required anymore, since existing node role label "worker" will be used for selecting the nodes.

alebedev87: As we discussed with @ayesha54, we'll keep the current logic of adding the custom nodeobservability role label. The nodeselector passed to `NobMC` has to be used in `List` nodes API call to filter the nodes which will be added with the custom label, not for the labeling.

api/v1alpha1/nodeobservability_types.go
```diff
@@ -29,8 +29,8 @@ const (
 
 // NodeObservabilitySpec defines the desired state of NodeObservability
 type NodeObservabilitySpec struct {
-	// Labels is map of key:value pairs that are used to match against node labels
-	Labels map[string]string `json:"labels,omitempty"`
+	// NodeSelector is map of key:value pairs that are used to match against node labels
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
alebedev87, line 33: Having empty `NodeSelector` doesn't have much sense anymore as we don't update all the workers.
I recommend to set it as `+kubebuilder:validation:Required` to enforce the user to specify some nodeselector.

config/samples/nodeobservability_v1alpha1_nodeobservability.yaml
```diff
@@ -3,6 +3,6 @@ kind: NodeObservability
 metadata:
   name: cluster
 spec:
-  labels:
+  nodeSelector:
```
alebedev87, line 6: Other sample files need to be updated too: `nodeobservability_v1alpha1_nodeobservability-all.yaml` and `nodeobservability_v1alpha1_nodeobservabilitymachineconfig.yaml`.

alebedev87:
@ayesha54 : you need to run `make bundle` on your local copy and commit/push all the diff produced by it. This will copy the CRD definition from `config/` to `bundle/` which will help to fix the `verify` and most probably `e2e` CI steps. The e2e step installs the bundle from the repository's `bundle/manifests/` files before the test is run.

pkg/operator/controller/machineconfig/nodes.go
```diff
@@ -90,7 +90,7 @@ func (r *MachineConfigReconciler) ensureReqNodeLabelExists(ctx context.Context)
 
 	updNodeCount := 0
 	nodeList := &corev1.NodeList{}
-	if err := r.listWorkerNodes(ctx, nodeList); err != nil {
+	if err := r.listLabeledWorkerNodes(ctx, nodeList); err != nil {
```
bharath-b-rh, line 93: listLabeledWorkerNodes and listNoObsLabeledNodes methods won't be required now right? listNodes would be enough I think.

bundle/manifests/nodeobservability.olm.openshift.io_nodeobservabilities.yaml
```diff
@@ -51,6 +51,7 @@ spec:
                 - crio-kubelet
                 type: string
             required:
+            - nodeSelector
```
bharath-b-rh, line 54: nodeSelector is mandatory?

alebedev87: [I asked](https://github.com/openshift/node-observability-operator/pull/83#discussion_r943935746) Ayesha to make it mandatory. I think it makes sense because we don't have a meaningful behavior in case the nodeSelector is not specified (only possible with optional).
What happens now (with this PR) if the node selector is not specified: agent's daemon gets scheduled on all the worker nodes as the nodeSelector is not there. MC and MCP are created but no nodes get labeled with `nodeObservability` role, so no CRIO pprof configuration is applied. Further `NodeObservabilityRun`s won't work as the pprof unix socket is not setup.

api/v1alpha1/nodeobservability_types.go
```diff
@@ -29,8 +29,9 @@ const (
 
 // NodeObservabilitySpec defines the desired state of NodeObservability
 type NodeObservabilitySpec struct {
-	// Labels is map of key:value pairs that are used to match against node labels
-	Labels map[string]string `json:"labels,omitempty"`
+	// +kubebuilder:validation:Required
+	// NodeSelector is map of key:value pairs that are used to match against node labels
```
alebedev87, line 33:
```suggestion
	// NodeSelector is map of key:value pairs that are used to match against node labels to be observed
```

pkg/operator/controller/machineconfig/nodes.go
```diff
@@ -249,14 +249,10 @@ func (r *MachineConfigReconciler) revertNodeUnlabeling(ctx context.Context) erro
...
 
-// listWorkerNodes returns the list of nodes having worker role label
-func (r *MachineConfigReconciler) listWorkerNodes(ctx context.Context, nodeList *corev1.NodeList) error {
-	workerNodeLabel := map[string]string{
-		WorkerNodeRoleLabelName: Empty,
-	}
-
-	if err := r.listNodes(ctx, nodeList, workerNodeLabel); err != nil {
-		return fmt.Errorf("failed to get the list of nodes with worker role: %w", err)
+// listLabeledWorkerNodes returns the list of labeled nodes with the node selector
```
alebedev87, line 252:
```suggestion
// listNodes returns the list of nodes selected by the currently reconciled machineconfig.
```

pkg/operator/controller/machineconfig/nodes.go
alebedev87, line 93:
```suggestion
	if err := r.listNodes(ctx, nodeList, r.CtrlConfig.Spec.NodeSelector); err != nil {
```
This would avoid the need of additional method.

api/v1alpha1/nodeobservabilitymachineconfig_types.go
```diff
@@ -26,6 +26,8 @@ import (
 // NodeObservabilityMachineConfigSpec defines the desired state of NodeObservabilityMachineConfig
 type NodeObservabilityMachineConfigSpec struct {
 	Debug NodeObservabilityDebug `json:"debug,omitempty"`
+	// +kubebuilder:validation:Required
+	NodeSelector map[string]string `json:"nodeSelector"`
```
alebedev87, line 30:
```suggestion
	// +kubebuilder:validation:Required
	// NodeSelector is a map of key:value pair that are used to match against node labels to be configured
	NodeSelector map[string]string `json:"nodeSelector"`
```

pkg/operator/controller/machineconfig/nodes.go
alebedev87: listLabeledWorkerNodes and listNoObsLabeledNodes methods are still needed because this PR doesn't change the logic of labeling the nodes to be observed, it just hooks into it changing which nodes should be labeled: only those from `.spec.nodeSelector`.

PR #87, opened 2022-09-01
CFE-510: Enforce the desired state of the operand objects

thejasn:
Description
===
Refactor reconcile loop logic to enforce required fields/schema of `DaemonSet`, `Service`, `ServiceAccount` and `NodeObservabilityMachineConfig` objects created by the operator. Simplifies error handling within the reconcile loop. 

## Changes
- `Dockerfile` - `podman` on v4.2 has short name aliasing is on by default and needs full image name to be present.
- `bundle/manifest` & `config/rbac/role.yaml` - Add `update` and `patch` RBAC verbs to enforced resources.
- `cmd/node-observability-operator/main.go` - Update kube-client to use non-cached client.
- `pkg/operator/controller/nodeobservability/simulate-error-schema.go` - Remove enforced objects from list, eventually need to remove `ErrTestObject` as well.
- `pkg/operator/controller/nodeobservability/controller_test.go` - Remove and fix tests logic after deleting usages of `ErrTestObject`.
- Enforces required fields of `DaemonSet`, `Service`, `ServiceAccount` and `NodeObservabilityMachineConfig` and adds additional tests.

Signed-off-by: Thejas N <thn@redhat.com>

### DIFF
--- a/pkg/operator/controller/nodeobservability/configmap.go
+++ b/pkg/operator/controller/nodeobservability/configmap.go
@@ -58,9 +58,9 @@ func (r *NodeObservabilityReconciler) createConfigMap(ctx context.Context, nodeO
 			if err := r.Create(ctx, configMap); err != nil {
 				return fmt.Errorf("failed to create target configmap %q: %w", configMapName, err)
 			}
+			r.Log.Info("created kubelet CA configmap", "configmap.namespace", configMapName.Namespace, "configmap.name", configMapName.Name)
 		}
 	}
 
-	r.Log.Info("created kubelet CA configmap", "configmap.namespace", configMapName.Namespace, "configmap.name", configMapName.Name)
 	return nil
 }
--- a/pkg/operator/controller/nodeobservability/configmap.go
+++ b/pkg/operator/controller/nodeobservability/configmap.go
@@ -28,9 +28,7 @@ func (r *NodeObservabilityReconciler) createConfigMap(ctx context.Context, nodeO
 	// Use the clusterWide client in order to get the configmap from openshift-config-managed namespace
 	// As the default client will only look for configmaps inside the namespace
 	if err := r.ClusterWideClient.Get(ctx, kbltCACMName, kbltCACM); err != nil {
-		if !errors.IsNotFound(err) {
-			return fmt.Errorf("error getting source configmap %q, %w", kbltCACMName, err)
-		}
+		return fmt.Errorf("error getting source configmap %q, %w", kbltCACMName, err)
 	}
 
 	// Copy the configmap into the operator namespace
@@ -49,19 +47,20 @@ func (r *NodeObservabilityReconciler) createConfigMap(ctx context.Context, nodeO
 	}
 
 	if err := controllerutil.SetControllerReference(nodeObs, configMap, r.Scheme); err != nil {
-		return fmt.Errorf("failed to set the controller reference for configmap %s/%s: %w", configMap.Namespace, configMap.Name, err)
+		return fmt.Errorf("failed to set the controller reference for target configmap %q: %w", configMapName, err)
 	}
 
-	if err := r.Get(ctx, configMapName, &corev1.ConfigMap{}); err == nil {
-		return nil
-	} else if !errors.IsNotFound(err) {
-		return fmt.Errorf("error getting target configmap %s, %w", configMapName, err)
+	if err := r.Get(ctx, configMapName, &corev1.ConfigMap{}); err != nil {
+		if !errors.IsNotFound(err) {
+			return fmt.Errorf("error getting target configmap %q, %w", configMapName, err)
+		} else {
+			// create configmap since it is not found
+			if err := r.Create(ctx, configMap); err != nil {
+				return fmt.Errorf("failed to create target configmap %q: %w", configMapName, err)
+			}
+		}
 	}
 
-	if err := r.Create(ctx, configMap); err != nil {
-		return fmt.Errorf("failed to create target configmap %s: %w", configMapName, err)
-	}
-
-	r.Log.Info("created configmap", "name", configMapName.Namespace, "name", configMapName.Name)
+	r.Log.Info("created kubelet CA configmap", "configmap.namespace", configMapName.Namespace, "configmap.name", configMapName.Name)
 	return nil
 }
--- a/pkg/operator/controller/nodeobservability/configmap.go
+++ b/pkg/operator/controller/nodeobservability/configmap.go
@@ -9,6 +9,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
 	v1alpha1 "github.com/openshift/node-observability-operator/api/v1alpha1"
 )
 
@@ -40,18 +42,14 @@ func (r *NodeObservabilityReconciler) createConfigMap(ctx context.Context, nodeO
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      configMapName.Name,
 			Namespace: configMapName.Namespace,
-			OwnerReferences: []metav1.OwnerReference{
-				{
-					Name:       nodeObs.Name,
-					Kind:       nodeObs.Kind,
-					UID:        nodeObs.UID,
-					APIVersion: nodeObs.APIVersion,
-				},
-			},
 		},
 		Data: map[string]string{
 			kbltCAMountedFile: kbltCACM.Data[kbltCAMountedFile],
 		},
+	}
+
+	if err := controllerutil.SetControllerReference(nodeObs, configMap, r.Scheme); err != nil {
+		return fmt.Errorf("failed to set the controller reference for configmap %s/%s: %w", configMap.Namespace, configMap.Name, err)
 	}
 
 	if err := r.Get(ctx, configMapName, &corev1.ConfigMap{}); err == nil {
--- a/pkg/operator/controller/nodeobservability/configmap.go
+++ b/pkg/operator/controller/nodeobservability/configmap.go
@@ -17,7 +17,7 @@ const (
 	srcKbltCAConfigMapNameSpace = "openshift-config-managed"
 )
 
-func (r *NodeObservabilityReconciler) createConfigMap(ctx context.Context, nodeObs *v1alpha1.NodeObservability, ns string) (bool, error) {
+func (r *NodeObservabilityReconciler) createConfigMap(ctx context.Context, nodeObs *v1alpha1.NodeObservability, ns string) error {
 	kbltCACM := &corev1.ConfigMap{}
 	kbltCACMName := types.NamespacedName{
 		Name:      srcKbltCAConfigMapName,
@@ -26,10 +26,9 @@ func (r *NodeObservabilityReconciler) createConfigMap(ctx context.Context, nodeO
 	// Use the clusterWide client in order to get the configmap from openshift-config-managed namespace
 	// As the default client will only look for configmaps inside the namespace
 	if err := r.ClusterWideClient.Get(ctx, kbltCACMName, kbltCACM); err != nil {
-		if errors.IsNotFound(err) {
-			return false, fmt.Errorf("unable to find source configMap %v: %w", kbltCACMName, err)
+		if !errors.IsNotFound(err) {
+			return fmt.Errorf("error getting source configmap %q, %w", kbltCACMName, err)
 		}
-		return false, fmt.Errorf("error getting source configMap %v, %w", kbltCACMName, err)
 	}
 
 	// Copy the configmap into the operator namespace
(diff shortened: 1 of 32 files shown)


pkg/operator/controller/nodeobservability/daemonset.go
```diff
@@ -208,3 +268,136 @@ func (r *NodeObservabilityReconciler) desiredDaemonSet(nodeObs *v1alpha1.NodeObs
...
+
+func indexedContainerEnv(envs []corev1.EnvVar) map[string]corev1.EnvVar {
+	indexed := make(map[string]corev1.EnvVar)
+	for _, e := range envs {
+		indexed[e.Name] = e
+	}
+	return indexed
+}
+
+func hasSecurityContextChanged(current, desired *corev1.SecurityContext) bool {
```
alebedev87, line 344: This function checks a bit more than the desired state really needs. It's more appropriate for a POD which got into the `restricted` POD security profile.

pkg/operator/controller/nodeobservability/daemonset.go
```diff
@@ -208,3 +268,136 @@ func (r *NodeObservabilityReconciler) desiredDaemonSet(nodeObs *v1alpha1.NodeObs
...
+
+	if desired.Capabilities != nil {
+		if current.Capabilities == nil {
+			return true
+		}
+
+		cmpCapabilities := cmpopts.SortSlices(func(a, b corev1.Capability) bool { return a < b })
+		if !cmp.Equal(desired.Capabilities.Add, current.Capabilities.Add, cmpCapabilities) {
+			return true
+		}
```
alebedev87, line 361: Same approach could have been followed for the `args` field to allow the reordered identical content.

alebedev87 (commented):
Part 1. Checked down to the daemonset. No tests reviewed.
To be continued...

pkg/operator/controller/nodeobservability/service.go
```diff
@@ -100,3 +142,38 @@ func (r *NodeObservabilityReconciler) desiredService(nodeObs *v1alpha1.NodeObser
...
+	copy(currentCopy, current)
+	sort.Sort(currentCopy)
+	desiredCopy := make(SortableServicePort, len(desired))
+	copy(desiredCopy, desired)
+	sort.Sort(desiredCopy)
+
+	for i := 0; i < len(currentCopy); i++ {
+		c := currentCopy[i]
+		d := desiredCopy[i]
+		if c.Name != d.Name || c.Port != d.Port || c.TargetPort.IntVal != d.TargetPort.IntVal {
```
alebedev87, line 174: Protocol is missing.

pkg/operator/controller/nodeobservability/serviceaccount.go
```diff
@@ -19,62 +21,54 @@ const (
...
 		if err := r.createServiceAccount(ctx, desired); err != nil {
-			return false, nil, err
+			return nil, err
 		}
+		r.Log.V(1).Info("successfully created serviceaccount", "sa.name", nameSpace.Name, "sa.namespace", nameSpace.Namespace)
 		return r.currentServiceAccount(ctx, nameSpace)
 	}
-	return true, current, err
+
+	return desired, r.Update(ctx, desired)
```
alebedev87, line 45: We do compare the owner references in the other operands. Let's be consistent.

alebedev87 (commented):
Part 2. Finished on the code. Will do the test later as they may change.

pkg/operator/controller/nodeobservability/configmap.go
```diff
@@ -41,30 +42,26 @@ func (r *NodeObservabilityReconciler) createConfigMap(ctx context.Context, nodeO
...
+
 	if err := r.Get(ctx, configMapName, &corev1.ConfigMap{}); err == nil {
-		return true, nil
+		return nil
 	} else if !errors.IsNotFound(err) {
-		return false, fmt.Errorf("error getting target configMap %s, %w", configMapName, err)
+		return fmt.Errorf("error getting target configmap %s, %w", configMapName, err)
 	}
 
 	if err := r.Create(ctx, configMap); err != nil {
```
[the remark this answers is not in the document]
thejasn: Yes, since that is being developed under a separate task.

pkg/operator/controller/nodeobservability/configmap.go
```diff
@@ -26,10 +28,9 @@ func (r *NodeObservabilityReconciler) createConfigMap(ctx context.Context, nodeO
 	// Use the clusterWide client in order to get the configmap from openshift-config-managed namespace
 	// As the default client will only look for configmaps inside the namespace
 	if err := r.ClusterWideClient.Get(ctx, kbltCACMName, kbltCACM); err != nil {
-		if errors.IsNotFound(err) {
-			return false, fmt.Errorf("unable to find source configMap %v: %w", kbltCACMName, err)
```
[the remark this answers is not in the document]
thejasn: Agreed, my mistake I missed the return during refactor.

pkg/operator/controller/nodeobservability/configmap.go
```diff
@@ -41,30 +42,26 @@ func (r *NodeObservabilityReconciler) createConfigMap(ctx context.Context, nodeO
...
 	}
 
 	if err := r.Create(ctx, configMap); err != nil {
-		return false, fmt.Errorf("failed to create target configMap %s: %w", configMapName, err)
+		return fmt.Errorf("failed to create target configmap %s: %w", configMapName, err)
 	}
 
-	r.Log.Info("created ConfigMap", "ConfigMap.Namespace", configMapName.Namespace, "ConfigMap.Name", configMapName.Name)
-	return true, nil
+	r.Log.Info("created configmap", "name", configMapName.Namespace, "name", configMapName.Name)
```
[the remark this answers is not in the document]
thejasn: I'm assuming this should be `created target configmap`. Right?

pkg/operator/controller/nodeobservability/daemonset.go
```diff
@@ -31,59 +34,114 @@ const (
...
 
 // currentDaemonSet check if the daemonset exists
-func (r *NodeObservabilityReconciler) currentDaemonSet(ctx context.Context, nameSpace types.NamespacedName) (bool, *appsv1.DaemonSet, error) {
+func (r *NodeObservabilityReconciler) currentDaemonSet(ctx context.Context, nameSpace types.NamespacedName) (*appsv1.DaemonSet, error) {
 	ds := &appsv1.DaemonSet{}
-	if err := r.Get(ctx, nameSpace, ds); err != nil || r.Err.Set[dsObj] {
-		if errors.IsNotFound(err) || r.Err.NotFound[dsObj] {
-			return false, nil, nil
-		}
-		if r.Err.Set[dsObj] {
```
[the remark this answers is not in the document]
thejasn: Once the rbac changes are pushed, we can delete this field. Currently there are references in `simulate-error-schema.go` and `controller_test.go`.

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -157,34 +158,28 @@ func (r *NodeObservabilityReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	// ensure service
-	haveSvc, svc, err := r.ensureService(ctx, nodeObs, r.Namespace)
+	svc, err := r.ensureService(ctx, nodeObs, r.Namespace)
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to ensure service : %w", err)
-	} else if !haveSvc {
-		return ctrl.Result{}, fmt.Errorf("failed to get service")
 	}
-	r.Log.V(1).Info("Service ensured", "Namespace", svc.Namespace, "Name", svc.Name)
+	r.Log.V(1).Info("service ensured", "svc.namespace", svc.Namespace, "svc.name", svc.Name)
```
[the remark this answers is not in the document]
thejasn: Yeah, I had the same thought. But turns out the actual log has multiple resources with the key as `name` and thees keyvalues are appended to the log. In the end, it looks something like this (I'll put the actual logs a little later), 
```bash
{"level":"debug","ts":"2022-09-13T09:58:12.129Z","logger":"controller.nodeobservability","msg":"service ensured","reconciler group":"nodeobservability.olm.openshift.io","reconciler kind":"NodeObservability","name":"cluster","namespace":"","svc.namespace":"node-observability-operator","svc.name":"node-observability-agent"}
```
Since these are debug logs, we could print the entire resource as well. Also this log formal needs to be changed, I'm thinking of introducing a separate log package which fixes the format. (need to move from json to text format)

pkg/operator/controller/nodeobservability/daemonset.go
thejasn: I agree, I'm fine either ways retaining (future proof) or strip it down to whats required.

pkg/operator/controller/nodeobservability/configmap.go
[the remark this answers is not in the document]
alebedev87: Well yes, it's the target confgimap, the separation in the logs was to distinguish the 2 configmaps to help the engineering. I thought that at this point (when the CA is copied over) we can already talk about "the configmap" or even about the "the kubelet CA configmap" as the later means a little more to the user.

pkg/operator/controller/nodeobservability/controller.go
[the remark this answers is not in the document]
alebedev87: Yes, true, I forgot that the logs are decorated with the request which is being reconciled. You are right let's keep this format.

> Since these are debug logs, we could print the entire resource as well. Also this log formal needs to be changed, I'm thinking of introducing a separate log package which fixes the format. (need to move from json to text format)

Well, the structured logging is what is the trend now, it's less human readable but better machine "parsable". As a matter of fact, you can switch in between the text and the JSON format by toggling `Developement` [field](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log/zap#Options) in [the zap options](https://github.com/openshift/node-observability-operator/blob/1f92243033393e4bfdb3598531df96c0f991063e/cmd/node-observability-operator/main.go#L87-L90).

pkg/operator/controller/nodeobservability/daemonset.go
alebedev87: It's tempting to keep this but I'd use [the Occam's razor principle](https://en.wikipedia.org/wiki/Occam%27s_razor) and would bring the needed things when they are needed.

pkg/operator/controller/nodeobservability/daemonset.go
```diff
@@ -31,59 +33,101 @@ const (
...
+
 		return r.currentDaemonSet(ctx, nameSpace)
 	}
-	return true, current, err
+
+	updated, err := r.updateDaemonset(ctx, current, desired)
+	if err != nil {
+		return nil, fmt.Errorf("failed to update the daemonset %q: %w", nameSpace, err)
+	}
+
```
alebedev87, line 66: Can you please add a log for the updated daemon set, similar to what you did after the creation?

pkg/operator/controller/nodeobservability/configmap.go
```diff
@@ -41,30 +40,27 @@ func (r *NodeObservabilityReconciler) createConfigMap(ctx context.Context, nodeO
...
+			// create configmap since it is not found
+			if err := r.Create(ctx, configMap); err != nil {
+				return fmt.Errorf("failed to create target configmap %q: %w", configMapName, err)
+			}
+		}
 	}
 
-	r.Log.Info("created ConfigMap", "ConfigMap.Namespace", configMapName.Namespace, "ConfigMap.Name", configMapName.Name)
-	return true, nil
+	r.Log.Info("created kubelet CA configmap", "configmap.namespace", configMapName.Namespace, "configmap.name", configMapName.Name)
```
alebedev87, line 64: This log has to be moved upper, to where the creation happens. Because if the target configmap is already there - no creation happens.

pkg/operator/controller/nodeobservability/daemonset_util.go
```diff
@@ -0,0 +1,240 @@
...
+	changed := false
+
+	updatedContainers := make([]corev1.Container, len(desired))
+	copy(updatedContainers, desired)
+
+	currentContMap := buildIndexedContainerMap(current)
+	desiredContMap := buildIndexedContainerMap(desired)
+
+	// let's check that all the current containers have the desired values set
+	for currName, currCont := range currentContMap {
```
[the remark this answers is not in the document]
thejasn: I've fixed the implementation already. But there was an issue with the associated test which checks for injected containers. Fixed that.

pkg/operator/controller/nodeobservability/daemonset_test.go
```diff
@@ -162,19 +591,319 @@ func TestEnsureDaemonset(t *testing.T) {
...
+				).build(),
+			expectedDaemonset: testDaemonset("daemonset", "test-namespace", "test-sa").
+				withContainers(testContainer("agent", "agent:v2").
+					withArgs([]string{"--arg=value"}...).
+					build(),
+				).build(),
+			expectUpdate: true,
+		},
+		{
+			name: "container injected into daemonset",
```
alebedev87, line 656: Can you please add a test case for injected volume/volumemounts too?

pkg/operator/controller/nodeobservability/service_test.go
```diff
@@ -20,73 +20,220 @@ import (
...
+			},
+			expectedService: testControllerService(
+				podName,
+				test.TestNamespace,
+				map[string]string{"app": "nodeobservability", "nodeobs_cr": "test"},
+				map[string]string{injectCertsKey: podName},
+			),
+		},
+		{
+			name: "existing service, extra annotations present",
```
alebedev87, line 175: Would be good to highlight on the fact the the annotation is accepted/kept. Because just by looking at the name, it's not clear whether the ensure function is supposed to remove or keep it.

pkg/operator/controller/nodeobservability/daemonset_test.go
thejasn: They are already present, changed the test name to make it more clear.

pkg/operator/controller/nodeobservability/service.go
```diff
@@ -28,46 +34,89 @@ var (
...
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to get service %s/%s due to: %w", nameSpace.Namespace, nameSpace.Name, err)
+	} else if err != nil && errors.IsNotFound(err) {
+
+		// creating service since it is not found
 		if err := r.createService(ctx, desired); err != nil {
-			return false, nil, err
+			return nil, err
 		}
+		r.Log.V(1).Info("successfully created service", "svc.name", nameSpace.Name, "svc.namespace", nameSpace.Namespace)
```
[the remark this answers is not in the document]
thejasn: I think this is already addressed.

PR #89, opened 2022-09-19
api: v1alpha2 version added. part 1

alebedev87:
Follow-up of https://github.com/openshift/node-observability-operator/pull/83. `NodeObservability.Spec.NodeSelector` field replaced the existing `NodeObservability.Spec.Labels` which broke the compatibility. This PR aims as adding a new API version `v1alpha2` which re-introduces the breaking API feature. [Kubebuilder guide](https://book.kubebuilder.io/multiversion-tutorial/tutorial.html) was followed.

**Things changed**

- `v1alpha2` version added by simply coping from `v1alpha1` at its state as per https://github.com/openshift/node-observability-operator/pull/83
- `v1alpha1` reverted to the state before https://github.com/openshift/node-observability-operator/pull/83
- Manifests updated (`v1alpha2` is the storage version for all the types)
- `main.go` is now in the root directory as per the new kubebuilder layout
- Controllers code migrated to `v1apha2` version everywhere
- Conversion logic added but not enabled in the manifests
- `AllNamespaces` install mode added to CSV. It will be necessary for the conversion webhook to work.

**Note**
- Conversion webhook cannot be added at this PR as it requires `AllNamespaces` installmode only. The Prow config is not yet ready for it, so this PR introduces both modes `OwnNamespace` and `AllNamespaces`. The `OwnNamespace` mode will be removed along with the addition of the conversion webhook in [the part2 PR](https://github.com/openshift/node-observability-operator/pull/91).
- e2e test and the convenience script for CA injection are coming in [the part2 PR](https://github.com/openshift/node-observability-operator/pull/91) too.


ayesha54:
great effort @alebedev87

main.go
```diff
@@ -64,20 +65,25 @@ var (
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 	utilruntime.Must(nodeobservabilityv1alpha1.AddToScheme(scheme))
+	utilruntime.Must(nodeobservabilityv1alpha2.AddToScheme(scheme))
 	utilruntime.Must(securityv1.AddToScheme(scheme))
 	utilruntime.Must(rbacv1.AddToScheme(scheme))
 	utilruntime.Must(mcv1.AddToScheme(scheme))
+	utilruntime.Must(nodeobservabilityv1alpha2.AddToScheme(scheme))
```
thejasn, line 72: Duplicate.

main.go
```diff
@@ -86,6 +92,7 @@ func main() {
 	flag.StringVar(&tokenFile, "token-file", defaultTokenFile, "The path of the service account token.")
 	flag.StringVar(&caCertFile, "ca-cert-file", defaultCACertFile, "The path of the CA cert of the Agents' signing key pair.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false, "Enable leader election for controller manager. "+"Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&enableWebhook, "enable-webhook", false, "Enable the webhook server(s). Defaults to true.")
```
thejasn, line 95:
```suggestion
	flag.BoolVar(&enableWebhook, "enable-webhook", true, "Enable the webhook server(s). Defaults to true.")
```

thejasn (changes requested):
Just a couple of things.

main.go
alebedev87: This was done on purpose, I didn't want to start webhook server until the conversion config is not there yet. Will do it [in the followup PR](https://github.com/openshift/node-observability-operator/pull/91).

thejasn: ahh, that makes sense.

PR #90, opened 2022-09-20
CFE-603: Enforce scc and move clusterrole to manifest

thejasn:
Description
---
Enforce scc created for `node-observability-agents` and move clusterrole creation to manifest to reduce operator permissions.

Changes
---
- `bundle/manifests`: Deleted and re-generated manifest. Removes stale resources from when they were manually added.
- `config/rbac/operator_role.yaml`: Add operand clusterrole to manifest.
- `config/rbac/role.yaml`: Reduce permissions unnecessary provided to the controller.
- `pkg/operator/controller/nodeobservability/configmap.go`: Added more logs for debugging.
- `pkg/operator/controller/nodeobservability/controller.go` and `pkg/operator/controller/nodeobservability/controller_test.go`
    - Reduce rbac permissions
    - Enforce scc resource creation
    - Verify clusterrole's existence before creating a clusterrolebinding
    - Fix unit tests
-  `pkg/operator/controller/nodeobservability/rbac.go` and `pkg/operator/controller/nodeobservability/rbac_test.go`
    - Remove ensuring logic for cluster role
    - Fix unit tests
    - Add verify cluster role function


pkg/operator/controller/nodeobservability/rbac.go
```diff
@@ -3,163 +3,81 @@ package nodeobservabilitycontroller
...
-			return false, nil, err
-		}
-		return r.currentClusterRole(ctx, nameSpace)
-	}
-	return true, current, err
-}
+	// TODO: standardize resource names to keep them compatible
+	clusterRoleName        = "node-observability-operator-controller-role"
+	clusterRoleBindingName = "node-observability-operator-controller-role-binding"
+)
```
thejasn, line 23: Currently dummy names are assigned. Need to look into how to name these resources. Currently has a lot of inconsistency, 
- the SA that is mounted on the nob agent is called `node-observability-sa`
- the clusterrole assigned to the agent used to be called `node-observability-cr`
- the clusterrolebinding used to be called `node-observability-crb` 
Even though all of these are created for the operand.

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -68,16 +64,13 @@ type NodeObservabilityReconciler struct {
 //+kubebuilder:rbac:urls=/node-observability-pprof,verbs=get;
 //+kubebuilder:rbac:groups=authentication.k8s.io,resources=tokenreviews,verbs=create;
 //+kubebuilder:rbac:groups=authorization.k8s.io,resources=subjectaccessreviews,verbs=create;
-//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles,verbs=list;get;create;watch;delete;
-//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings,verbs=list;get;create;watch;delete;
-//+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=list;get;create;watch;use;delete;
-//+kubebuilder:rbac:groups=apps,namespace=node-observability-operator,resources=daemonsets,verbs=list;get;create;watch;update;patch;
-//+kubebuilder:rbac:groups=core,namespace=node-observability-operator,resources=secrets,verbs=list;get;create;watch;delete;
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles,verbs=get,resourceNames=node-observability-operator-controller-role
```
alebedev87, line 67: I believe you need to add `bind` verb too, to restrict the possibility to bind only to the resource `node-observability-operator-controller-role`.

pkg/operator/controller/nodeobservability/scc.go
```diff
@@ -66,39 +71,42 @@ func (r *NodeObservabilityReconciler) desiredSecurityContextConstraints(nodeObs
...
+		Groups:                   []string{"system:cluster-admins", "system:nodes"},
 	}
 	return scc
 }
 
+func (r *NodeObservabilityReconciler) updateSecurityContextConstraintes(ctx context.Context, current, desired *securityv1.SecurityContextConstraints) (bool, error) {
+	opts := cmpopts.IgnoreFields(securityv1.SecurityContextConstraints{}, "TypeMeta", "Users", "SeccompProfiles", "AllowedUnsafeSysctls", "ForbiddenSysctls")
+	if !cmp.Equal(current, desired, opts) {
+		// copy over current object metadata to updated
+		updated := desired.DeepCopy()
```
alebedev87, line 100:
```suggestion
		updated := current.DeepCopy()
```

pkg/operator/controller/nodeobservability/scc.go
```diff
@@ -66,39 +71,42 @@ func (r *NodeObservabilityReconciler) desiredSecurityContextConstraints(nodeObs
...
+		updated.ObjectMeta = current.ObjectMeta
```
alebedev87, line 101: 1) We should copy from the desired
2) `ObjectMeta` was the only thing not different in current and desired. We have to copy all the fields one by one. In the case the `cmp` package maybe not not even needed.

pkg/operator/controller/nodeobservability/scc.go
```diff
@@ -66,39 +71,42 @@ func (r *NodeObservabilityReconciler) desiredSecurityContextConstraints(nodeObs
...
+		ObjectMeta:               metav1.ObjectMeta{Name: sccName},
+		Priority:                 &priority,
 		AllowPrivilegedContainer: true,
-		AllowHostIPC:             false,
+		DefaultAddCapabilities:   nil,
+		RequiredDropCapabilities: []corev1.Capability{"MKNOD"},
+		AllowedCapabilities:      nil,
+		AllowHostDirVolumePlugin: true,
+		Volumes:                  []securityv1.FSType{securityv1.FSTypeHostPath, securityv1.FSTypeSecret, securityv1.FSTypeConfigMap},
 		AllowHostNetwork:         true,
```
alebedev87, line 82:
```suggestion
		AllowHostNetwork:         false,
```
We don't use hostNetwork for now.

pkg/operator/controller/nodeobservability/scc.go
```diff
@@ -66,39 +71,42 @@ func (r *NodeObservabilityReconciler) desiredSecurityContextConstraints(nodeObs
...
-			Type: securityv1.SupplementalGroupsStrategyRunAsAny,
-		},
-		Volumes: []securityv1.FSType{securityv1.FSTypeHostPath, securityv1.FSTypeSecret, securityv1.FSTypeConfigMap},
+		Groups:                   []string{"system:cluster-admins", "system:nodes"},
 	}
 	return scc
 }
 
+func (r *NodeObservabilityReconciler) updateSecurityContextConstraintes(ctx context.Context, current, desired *securityv1.SecurityContextConstraints) (bool, error) {
+	opts := cmpopts.IgnoreFields(securityv1.SecurityContextConstraints{}, "TypeMeta", "Users", "SeccompProfiles", "AllowedUnsafeSysctls", "ForbiddenSysctls")
```
alebedev87, line 97: Why these are filtered out: `"SeccompProfiles", "AllowedUnsafeSysctls", "ForbiddenSysctls"` ?

bundle/manifests/node-observability-operator-controller-role_rbac.authorization.k8s.io_v1_clusterrole.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  labels:
+    scc: node-observability-scc
+  name: node-observability-operator-controller-role
```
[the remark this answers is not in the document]
thejasn: Thoughts on https://github.com/openshift/node-observability-operator/pull/90#discussion_r976435550? A few other things need to be renamed as well.
This does break backward compatibility in terms of clean-up.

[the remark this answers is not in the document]
alebedev87: Indeed, this is not very clean as we leave some leftover for the installations which have already been made.
However keeping misleading or not optimal naming is not clean either.
Since we are in TechPreview and I doubt the big number of customers for this operator, I'd say let's make the clear naming. Also, if we will do so for all the 3 (service account, cluster role and cluster role binding), this will have the least possible impact as the cluster role binding will be pointing to a not existing service account.

pkg/operator/controller/nodeobservability/rbac_test.go
```diff
@@ -280,27 +168,19 @@ func TestEnsureClusterRole(t *testing.T) {
...
+			opts := cmpopts.EquateEmpty()
+			if diff := cmp.Diff(tc.expectedCRB.Subjects, crb.Subjects, opts); diff != "" {
+				t.Fatalf("clusterrolebinding subjects mismatch: \n%s", diff)
 			}
-			if gotExist != tc.expectedExist {
-				t.Errorf("expected service account's exist to be %t, got %t", tc.expectedExist, gotExist)
+
+			if diff := cmp.Diff(tc.expectedCRB.RoleRef, crb.RoleRef, opts); diff != "" {
+				t.Fatalf("clusterrolebinding roleref mismatch: \n%s", diff)
 			}
```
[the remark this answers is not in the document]
thejasn: We would have to ignore a bunch of fields (type metadata, object metadata, etc), would you prefer the ignore or this?

pkg/operator/controller/nodeobservability/scc.go
thejasn: These fields were previously unset and looks like we are just relying on the defaults. Do you want me to explicitly add them to desired? (at least for the ones that have values as defaults)

pkg/operator/controller/nodeobservability/controller.go
alebedev87: Just checked this, actually `bind` verb is only needed when the privileged escalation needs to be bypassed. If your code is working (manged to create the clusterrolebinding for the agent role) I think it's not needed then.

pkg/operator/controller/nodeobservability/scc.go
alebedev87: I think that relying on the defaults makes sense for "workload resources" (PODs, deployments, etc.). SCC is a sort of security framework which will put boundaries on what the agent can do. I think we should be precise in every field of this framework.
If we don't set some fields than they should remain empty because otherwise the customer can update them: add allowed unsafe/forbidded sysctls, etc. We should ask this question to ourselves: do we want the agent to be added with unsafe/forbidden syscalls? I'd say no unless there is a strict necessity.
Although `Users` and `Groups`, I'd keep them ignored to let the customers to add other SAs to this SCC.

bundle/manifests/node-observability-operator-controller-role_rbac.authorization.k8s.io_v1_clusterrole.yaml
```diff
@@ -0,0 +1,57 @@
...
+  verbs:
+  - get
+  - list
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - get
+  - list
```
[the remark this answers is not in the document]
thejasn: pod permissions weren't required, but the nodes permission are needed since we make a http call to trigger debugging. 
```
time="2022-10-05T10:29:26Z" level=error msg="errors encountered while running Kubelet:\nerror with HTTP request for kubelet profiling https://10.0.158.202:10250/debug/pprof/profile: statusCode 403\n" module=handler

```

bundle/manifests/node-observability-operator-controller-role_rbac.authorization.k8s.io_v1_clusterrole.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  labels:
+    scc: node-observability-scc
```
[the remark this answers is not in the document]
thejasn: Seems unnecessary, kind of was being used before. Might have just been a way to denote the cluster role was linked with the said SCC.

bundle/manifests/node-observability-operator.clusterserviceversion.yaml
```diff
@@ -321,17 +321,17 @@ spec:
           - delete
           - get
           - list
+          - patch
+          - update
           - watch
         - apiGroups:
           - rbac.authorization.k8s.io
+          resourceNames:
+          - node-observability-operator-controller-role
```
alebedev87, line 330:
```suggestion
          - node-observability-operator-agent
```

config/rbac/operand_role.yaml
```diff
@@ -0,0 +1,49 @@
...
+metadata:
+  creationTimestamp: null
+  name: agent
+rules:
+- apiGroups:
+  - "security.openshift.io"
+  resources:
+  - securitycontextconstraints
+  resourceNames:
+  - node-observability-scc
```
alebedev87, line 13: Would be good to remove the trailing `-scc` from the SCC name as it's redundant.

hack/clean-up.sh
```diff
@@ -2,7 +2,7 @@
 oc patch nodeobservability/nodeobservability-sample --patch '{"metadata":{"finalizers": []}}' --type=merge
 oc delete nodeobservability nodeobservability-sample
 oc delete scc node-observability-scc
-oc delete clusterrole node-observability-cr
-oc delete clusterrolebinding node-observability-crb
+oc delete clusterrole node-observability-controller-role
+oc delete clusterrolebinding node-observability-controller-role-binding
```
alebedev87, line 6:
```suggestion
oc delete clusterrole node-observability-operator-agent
oc delete clusterrolebinding node-observability-agent
```

pkg/operator/controller/nodeobservability/rbac.go
```diff
@@ -3,163 +3,78 @@ package nodeobservabilitycontroller
...
+		r.Log.Info("created clusterrolebinding", "clusterrolebinding.Name", clusterRoleBindingName)
+		return r.currentClusterRoleBinding(ctx)
 	}
-	return true, current, err
+
+	if _, err = r.updateClusterRoleBinding(ctx, current, desired); err != nil {
+		return nil, fmt.Errorf("failed to update clusterrolebinding %s due to: %w", clusterRoleBindingName, err)
+	}
+
+	return r.currentClusterRoleBinding(ctx)
```
alebedev87, line 63: Can you please do the same as in the daemonset controller?

- return `true` if the clusterrolebinding was updated and call `currentClusterRoleBinding` and log the `updated` message
- return `current` variable otherwise

pkg/operator/controller/nodeobservability/scc.go
```diff
@@ -14,50 +15,53 @@ import (
...
 	}
-	return true, current, err
+
+	updated, err := r.updateSecurityContextConstraintes(ctx, current, desired)
+	if err != nil {
+		return nil, fmt.Errorf("failed to update securitycontextconstraints %q due to: %w", sccName, err)
+	}
+
+	if updated {
+		r.Log.V(1).Info("updated securitycontextconstraints", "scc.Name", sccName)
```
alebedev87, line 45:
```suggestion
		r.Log.V(1).Info("updated securitycontextconstraints", "scc.name", sccName)
```

bundle/manifests/node-observability-operator-controller-role_rbac.authorization.k8s.io_v1_clusterrole.yaml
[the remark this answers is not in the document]
alebedev87: Yes, true,[ kubelet checks the node](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/server/auth.go#L60) permissions to allow the access to it API. Although, I doesn't seem to need the `nodes` resource, only `nodes/proxy`, `list` verb doesn't seem to be needed either.

pkg/operator/controller/nodeobservability/daemonset.go
```diff
@@ -25,7 +25,7 @@ const (
 	kbltCAMountedFile = "ca-bundle.crt"
 	kbltCAName        = "kubelet-ca"
 	defaultScheduler  = "default-scheduler"
-	daemonSetName     = "node-observability-ds"
+	daemonSetName     = "node-observability-agent"
```
alebedev87, line 28: I'm 100% for this change, it should have been named this way from the beginning.
But I'm a little worried about the migration. What will happen after the new NOB operator version will be rolled out? Will we have 2 daemons? How this will be handled at the profiling time when the operator will need to send a request to each agent?

hack/clean-up.sh
thejasn: I think we can delete this script, seems unused.

pkg/operator/controller/nodeobservability/daemonset.go
thejasn: I did this assuming we are ok due to this being in tech-preview. 
https://github.com/openshift/node-observability-operator/pull/90#discussion_r985533378

Maybe we mention in release notes, that we need to uninstall previous and cannot directly upgrade. Seems cleaner than adding migration code for tech-preview.

alebedev87: As discussed, let's make this experience as smooth as possible for the customers, especially knowing that the cleanup is feasible and should not take much code. Asking for the reinstall is indeed a measure and we can always use it but I'd prefer to keep it for the case when it's hardly avoidable which is far from this situation.
The rest of the renamings (roles,SCCs) don't have a big impact: no compute resources used, no rights granted. So, I propose to let them rot (=no migration implemented).

PR #91, opened 2022-09-20
api: v1alpha2 version added. part 2

alebedev87:
Followup of https://github.com/openshift/node-observability-operator/pull/89:
- Conversion webhook in the manifests enabled
- Convenience script to setup cluster certificate for the conversion webhook added
- e2e test scenario with `v1alpha1` version added:
    - changes put into a separate commit
    - 2 `nob`s are created now, so the cleanup has to be reliable, that's what the most of the code is doing
    - e2e now expects the operator to be available
    - more logs added
- samples updated to be more specific on `nodeSelector`
- `debug` is set as default log level, since it's still quite a young and often unstable operator

### DIFF
--- a/config/samples/nodeobservability_v1alpha2_nodeobservability.yaml
+++ b/config/samples/nodeobservability_v1alpha2_nodeobservability.yaml
@@ -4,5 +4,5 @@ metadata:
   name: cluster
 spec:
   nodeSelector:
-    node-role.kubernetes.io/worker: ""
+    kubernetes.io/hostname: worker-a
   type: crio-kubelet
--- a/test/e2e/nodeobservability_test.go
+++ b/test/e2e/nodeobservability_test.go
@@ -20,15 +20,23 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	operatorv1alpha1 "github.com/openshift/node-observability-operator/api/v1alpha1"
 	operatorv1alpha2 "github.com/openshift/node-observability-operator/api/v1alpha2"
 )
 
 var (
 	ctx = context.Background()
+)
+
+const (
+	hostnameLabel   = "kubernetes.io/hostname"
+	workerRoleLabel = "node-role.kubernetes.io/worker"
 )
 
 var _ = Describe("Node Observability Operator end-to-end test suite", Ordered, func() {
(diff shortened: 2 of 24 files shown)


test/e2e/nodeobservability_test.go
```diff
@@ -134,13 +142,122 @@ var _ = Describe("Node Observability Operator end-to-end test suite", Ordered, f
...
+		if !CurrentSpecReport().Failed() {
+			GinkgoWriter.Println("Cleaning up NOB")
+			Expect(k8sClient.Delete(ctx, nodeobservability)).To(Succeed())
+			GinkgoWriter.Println("Waiting for NOBMC to disappear")
+			nobmc := &operatorv1alpha1.NodeObservabilityMachineConfig{}
+			Eventually(func() bool {
+				return errors.IsNotFound(k8sClient.Get(ctx, types.NamespacedName{Name: "cluster"}, nobmc))
+			}, 600, time.Second).Should(BeTrue())
+		} else {
+			// keep NOB CR to preserve all the dependent objects
```
thejasn, line 256: I was thinking of doing this by running a must-gather within out test job. But I guess this works as well.

config/samples/nodeobservability_v1alpha2_nodeobservability.yaml
```diff
@@ -4,5 +4,5 @@ metadata:
   name: cluster
 spec:
   nodeSelector:
-    node-role.kubernetes.io/worker: ""
+    kubernetes.io/hostname: worker-a
```
thejasn, line 7: I'm guessing this was unintentional? Since worker nodes don't actually get this label. 
```bash
ip-10-0-171-177.ec2.internal   Ready    worker   3h30m   v1.24.0+1ff0731   beta.kubernetes.io/arch=amd64,beta.kubernetes.io/instance-type=m6i.xlarge,beta.kubernetes.io/os=linux,failure-domain.beta.kubernetes.io/region=us-east-1,failure-domain.beta.kubernetes.io/zone=us-east-1c,kubernetes.io/arch=amd64,kubernetes.io/hostname=ip-10-0-171-177.ec2.internal,kubernetes.io/os=linux,node-role.kubernetes.io/worker=,node.kubernetes.io/instance-type=m6i.xlarge,node.openshift.io/os_id=rhcos,topology.ebs.csi.aws.com/zone=us-east-1c,topology.kubernetes.io/region=us-east-1,topology.kubernetes.io/zone=us-east-1c

```

test/e2e/nodeobservability_test.go
alebedev87: It worked kinda OK when was run on local: allowed to keep NodeObservability CRs and all owned objects. In the CI, it's a bit less useful as `must-gather` [doesn't seem to collect non CVO resources](https://github.com/openshift/must-gather#collection-scripts). However `MCP` and `MC` which are owned by `NOBMC` get saved in CI too. In this regard I'm thinking whether running `must-gather` from the job would bring any more data than the "standard" must-gather done after the test.

config/samples/nodeobservability_v1alpha2_nodeobservability.yaml
alebedev87: The value of the label is irrelevant, it's just an example. I didn't want it to look as close as possible to the real values. I just wanted to show a sample which would be more granular than "all workers".
However on GCP the hostname label does contain `worker-a` substring.

thejasn: Got it. I generally use the samples during testing (I guess this is going to be an extra step). But the expected key on GCP is not `kubernetes.io/hostname:`. Right?

test/e2e/nodeobservability_test.go
thejasn: I was thinking of just re-using steps that upload the must-gathers as artifacts and providing must-gather a custom collection script from our operator repo (we can pull all the resources we need collected).

config/samples/nodeobservability_v1alpha2_nodeobservability.yaml
alebedev87: Yes, the expected key is `kubernetes.io/hostname`, it's present on all nodes on GCP and AWS at least.

alebedev87:
/label px-approved

PX approval for the new version has already been given in [the previous PR](https://github.com/openshift/node-observability-operator/pull/89).

PR #95, opened 2022-09-23
related agent image: use dedicated quay.io organization

alebedev87:
The agent image is now mirrored into the dedicated organization: https://quay.io/organization/node-observability-operator.


alebedev87:
/label docs-approved
/label qe-approved
/label px-approved

Operand image doesn't have functional changes (only OWNER file was updated).

PR #97, opened 2022-09-30
nob controller: daemonset test - refactor helpers

alebedev87:
Make test daemon set builder more practical which results into less chatty test cases.


thejasn:
Thanks, this seems slightly more specific by much more readable. 
/lgtm

alebedev87:
/label docs-approved
/label px-approved
/label qe-approved

Unit test helpers updates only: no need to any approvals by peer review

PR #99, opened 2022-10-07
CFE-599: Fix machineconfigcontroller flakiness

thejasn:
Description
---
Fixes `machineconfig` controller's reconciling dependency on `LastReconciled` time and makes it more simpler to debug. Also fixes issue in cleanup procedure of `nobmc` by relying on both `mcp` status and  `mcp` machine counts during cleanup/update/rollout. 

Removes reverting flow for handling failures during enabling CRI-O profiling. Now upon failures the user is able to see current status of `nobmc` and take action accordingly (delete `nobmc`, etc). 

### Change Log
---
- `pkg/operator/controller/machineconfig/controller.go`
    - Delete `CtrlConfig` from reconciler struct, attempts to state manage `nobmc`
    - Remove `LastReconciled` time check for kicking out reconcile requests
    - Simplify `addFinalizer()` and `removeFinalizer()`
    - Make `cleanUp()` and requeuing logic idempotent
    - Delete `monitorProgress()`, was redundant and was only added to handle multiple reverting flows (which are removed)
    - Delete `revertEnabledProfConf()`
- `pkg/operator/controller/machineconfig/crio.go`: Moved required constants to file and removed extra functions.
- `pkg/operator/controller/machineconfig/mcp.go`: Re-write logic to include machine counts for checks and rename, 
    - `checkNodeObservabilityMCPStatus()` ->  `syncNodeObservabilityMCPStatus()`
        - Also removes `revertEnabledProfConf()` flow. 
    - `checkWorkerMCPStatus()` -> `syncWorkerMCPStatus()`
-  `pkg/operator/controller/machineconfig/nodes.go`
    - Removed redundant functions for `patch` calls [`newAddPatch()`,`newRemovePatch()`,`getPatchValue()`]
- `pkg/util/slice/slice.go`: Adds new package as helpers for working with finalizers.


pkg/operator/controller/machineconfig/nodes.go
```diff
@@ -25,109 +25,128 @@ import (
...
-// listNodes returns the list of nodes matching the labels
-func (r *MachineConfigReconciler) listNodes(ctx context.Context, nodeList *corev1.NodeList, matchLabels map[string]string) error {
-	listOpts := []client.ListOption{
-		client.MatchingLabels(matchLabels),
+// ensureProfConfDisabled disables the profiling on the requested nodes by removing the nodeobservability label.
+func (r *MachineConfigReconciler) ensureProfConfDisabled(ctx context.Context, nomc *v1alpha2.NodeObservabilityMachineConfig) error {
+
+	nodeList, err := r.listNodes(ctx, map[string]string{NodeObservabilityNodeRoleLabelName: empty})
+	if err != nil {
+		return fmt.Errorf("failed to get the list of nodes labelled nodeobservability: %w", err)
```
thejasn, line 144: According to previous implementation, we seem to have hard coded the label selector used while disabling CRI-O profiling whereas during enabling we seem to taking it as user input. This doesn't feel right, I think we should enable and disable using the user given labels only. WDYT?

pkg/operator/controller/machineconfig/controller.go
```diff
@@ -80,228 +86,138 @@ func New(mgr ctrl.Manager) *MachineConfigReconciler {
...
-
-		if err := r.ClientUpdate(ctx, withFinalizers); err != nil {
-			return err
-		}
-
-		return nil
-	})
-
-	return withFinalizers, err
+	return ctrl.Result{}, nil
```
alebedev87, line 178: Would be nice to add a log message just before returning.

pkg/operator/controller/machineconfig/crio.go
```diff
@@ -25,54 +25,92 @@ import (
...
+	if err := ctrlutil.SetControllerReference(nomc, criomc, r.Scheme); err != nil {
 		return fmt.Errorf("failed to update controller reference in crio profiling machine config: %w", err)
 	}
 
 	if err := r.ClientCreate(ctx, criomc); err != nil && !errors.IsAlreadyExists(err) {
 		return fmt.Errorf("failed to create crio profiling machine config: %w", err)
 	}
 
-	r.Log.V(1).Info("Successfully created MachineConfig to enable CRI-O profiling", "CrioProfilingConfigName", CrioProfilingConfigName)
+	// TODO: check if there is a diff between desired and exists
```
alebedev87, line 69:
```suggestion
	// TODO: check if there is a diff between desired and existing
```

pkg/operator/controller/machineconfig/crio.go
```diff
@@ -25,54 +25,92 @@ import (
...
+
+	r.Log.V(1).Info("Successfully created MachineConfig to enable CRI-O profiling", "CrioProfilingConfigName", crioProfilingConfigName)
```
alebedev87, line 71:
```suggestion
	r.Log.V(1).Info("successfully created MachineConfig to enable CRI-O profiling", "mc.name", crioProfilingConfigName)
```

pkg/operator/controller/machineconfig/crio.go
```diff
@@ -25,54 +25,92 @@ import (
...
 	}
 
-	if err := r.ClientDelete(ctx, criomc); err != nil {
-		return fmt.Errorf("failed to remove crio profiling machine config: %w", err)
+	if err := r.ClientDelete(ctx, criomc); err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("failed to remove crio profiling machineconfig: %w", err)
 	}
 
-	r.Log.V(1).Info("Successfully removed MachineConfig to disable CRI-O profiling", "CrioProfilingConfigName", CrioProfilingConfigName)
+	r.Log.V(1).Info("successfully removed machineconfig to disable CRI-O profiling", "CrioProfilingConfigName", crioProfilingConfigName)
```
alebedev87, line 89:
```suggestion
	r.Log.V(1).Info("successfully removed machineconfig to disable CRI-O profiling", "mc.name", crioProfilingConfigName)
```

pkg/operator/controller/machineconfig/nodes.go
```diff
@@ -25,109 +25,128 @@ import (
...
-		r.Log.V(1).Info("Nodeobservability role is present on all the nodes with worker role", "ExistingNodeCount", existingNodeCount, "UpdatedNodeCount", updNodeCount)
-		return true, nil
+	if nodeCount == requiredNodeCount {
+		r.Log.V(1).Info("nodeobservability role is present on all the nodes with worker role", "UpdatedNodeCount", nodeCount)
+		return nil
 	}
 
-	return false, nil
+	r.Log.V(1).Info("nodeobservability role is not present on all the nodes with worker role", "Nodes", requiredNodeCount, "UpdatedNodeCount", nodeCount)
+	return fmt.Errorf("failed to ensure required label, expected updates on %d nodes but got %d", requiredNodeCount, nodeCount)
```
alebedev87, line 64: I believe the error will be returned back to `controller-runtime` which means that it'll be logged, making the log before redundant.

pkg/operator/controller/machineconfig/nodes.go
```diff
@@ -25,109 +25,128 @@ import (
...
-		if err := r.ClientPatch(ctx, &nodeList.Items[i], client.RawPatch(types.JSONPatchType, patch)); err != nil {
-			return false, err
+		err := r.patchNodeLabels(ctx, &nodeList.Items[i], patchAdd, ResourceLabelsPath, map[string]interface{}{NodeObservabilityNodeRoleLabelName: empty})
+		if err != nil {
+			return fmt.Errorf("failed to patch node due to %w", err)
 		}
-		r.Log.V(1).Info("Successfully added label", "Node", node.Name, "Label", NodeObservabilityNodeRoleLabelName)
-
-		updNodeCount++
+		r.Log.V(1).Info("successfully performed add operation for labels", "Node", node.Name, "Label", NodeObservabilityNodeRoleLabelName)
```
alebedev87, line 54:
```suggestion
		r.Log.V(1).Info("successfully performed add operation for labels", "node.name", node.Name, "node.label", NodeObservabilityNodeRoleLabelName)
```

pkg/operator/controller/machineconfig/nodes.go
```diff
@@ -25,109 +25,128 @@ import (
...
+		nodeCount++
 	}
 
-	if (existingNodeCount == len(nodeList.Items)) || (updNodeCount == len(nodeList.Items)) {
-		r.Log.V(1).Info("Nodeobservability role is present on all the nodes with worker role", "ExistingNodeCount", existingNodeCount, "UpdatedNodeCount", updNodeCount)
-		return true, nil
+	if nodeCount == requiredNodeCount {
+		r.Log.V(1).Info("nodeobservability role is present on all the nodes with worker role", "UpdatedNodeCount", nodeCount)
```
alebedev87, line 59:
```suggestion
		r.Log.V(1).Info("nodeobservability role is present on all the nodes with worker role", "updatednodecount", nodeCount)
```

pkg/operator/controller/machineconfig/nodes.go
```diff
@@ -25,109 +25,128 @@ import (
...
+// CRI-O profiling is applied.
+func (r *MachineConfigReconciler) ensureProfConfEnabled(ctx context.Context, nomc *v1alpha2.NodeObservabilityMachineConfig) error {
+
+	nodeList, err := r.listNodes(ctx, nomc.Spec.NodeSelector)
+	if err != nil {
+		return err
+	}
+
+	if len(nodeList.Items) == 0 {
+		r.Log.V(1).Info("No nodes matching the given selector were found", "NodeSelector", nomc.Spec.NodeSelector)
```
alebedev87, line 119:
```suggestion
		r.Log.V(1).Info("no nodes matching the given selector were found", "nodeselector", nomc.Spec.NodeSelector)
```

pkg/operator/controller/machineconfig/nodes.go
```diff
@@ -25,109 +25,128 @@ import (
...
+
+	err = r.ensureReqNodeLabelExists(ctx, nodeList)
+	if err != nil {
+		return fmt.Errorf("failed to ensure nodes are labelled: %w", err)
 	}
 
-	if err := r.listNodes(ctx, nodeList, nodeLabel); err != nil {
-		return fmt.Errorf("failed to get the list of nodes labeled nodeobservability: %w", err)
+	if err := r.enableCrioProf(ctx, nomc); err != nil {
+		return fmt.Errorf("failed to enable CRIO mc due to: %w", err)
```
alebedev87, line 129:
```suggestion
		return fmt.Errorf("failed to enable CRIO mc: %w", err)
```

pkg/operator/controller/machineconfig/nodes.go
```diff
@@ -25,109 +25,128 @@ import (
...
+	}
+
+	if err := r.createProfMCP(ctx, nomc); err != nil {
+		return fmt.Errorf("failed to create profiling mcp due to: %w", err)
```
alebedev87, line 133:
```suggestion
		return fmt.Errorf("failed to create profiling mcp: %w", err)
```

pkg/operator/controller/machineconfig/nodes.go
alebedev87: At the time of the disabling those are the same label selector as we add nob role label only to the nodes selected by the user provided label selector. I believe we only use the user provided one when we label nodes, in all the other cases (here and in nob `MCP`) we use the created label.

pkg/operator/controller/machineconfig/nodes.go
```diff
@@ -139,50 +158,30 @@ func escape(key string) string {
...
 
-// newPatch returns the patch data in the format required by the client
-func newPatch(op patchOp, pathPrefix string, patch map[string]interface{}) ([]byte, error) {
-	if patch == nil {
-		return nil, fmt.Errorf("patch data cannot be empty")
+// patchNodeLabels creates and patches the given node with the provided labels
+func (r *MachineConfigReconciler) patchNodeLabels(ctx context.Context, node *corev1.Node, op string, pathPrefix string, labels map[string]interface{}) error {
+	if labels == nil {
+		return fmt.Errorf("patch data cannot be empty")
 	}
```
alebedev87, line 165: Is this ever used? All calls of this function are guaranteed to get the labels.

pkg/operator/controller/machineconfig/nodes.go
```diff
@@ -139,50 +158,30 @@ func escape(key string) string {
...
 
-// newAddPatch returns the patch data to add in the format required by the client
-func newAddPatch(pathPrefix string, patch map[string]interface{}) []byte {
-	values := make([]ResourcePatchValue, 0, len(patch))
-	for k, v := range patch {
-		ppath := path.Join(pathPrefix, escape(k))
-		values = append(values, getPatchValue("add", ppath, v))
+	data, err := json.Marshal(values)
+	if err != nil {
+		return err
```
alebedev87, line 179:
```suggestion
		return fmt.Errorf("unable to marshal patch values: %w", err)
```

pkg/operator/controller/machineconfig/nodes.go
```diff
@@ -139,50 +158,30 @@ func escape(key string) string {
...
-
-// newRemovePatch returns the patch data to remove in the format required by the client
-func newRemovePatch(pathPrefix string, patch map[string]interface{}) []byte {
-	values := make([]ResourcePatchValue, 0, len(patch))
-	for k, v := range patch {
-		ppath := path.Join(pathPrefix, escape(k))
-		values = append(values, getPatchValue("remove", ppath, v))
+	if err := r.ClientPatch(ctx, node, client.RawPatch(types.JSONPatchType, data)); err != nil {
+		return err
 	}
```
alebedev87, line 184:
```suggestion
	return r.ClientPatch(ctx, node, client.RawPatch(types.JSONPatchType, data))
```

alebedev87 (commented):
Thanks for the PR!
Mostly logs and error handling plus the status update.

Another thing I wanted to bring up is the e2e test. I was thinking whether nobmc controller should check whether the crio profiling is already there or not. I think that the final answer is yes. However not checking this gives us a good chance to simply test the controller - we can just created `nobmc` CR and see the controller working from the beginning to the end.

pkg/operator/controller/machineconfig/controller.go
```diff
@@ -311,105 +227,52 @@ func (r *MachineConfigReconciler) updateStatus(ctx context.Context) error {
...
+	err := r.ensureProfConfDisabled(ctx, nomc)
 	if err != nil {
-		return true, fmt.Errorf("failed to ensure nodes are labelled: %w", err)
-	}
-	if !labelEnsured {
-		// not all (or none of) the labels are present,
-		// requeue to retry later
-		return true, nil
+		r.Log.V(1).Error(err, "failed to disable CRI-o profiling, retrying")
+		nomc.Status.SetCondition(v1alpha2.DebugEnabled, metav1.ConditionFalse, v1alpha2.ReasonInProgress, "disabling CRIO profiling in progress")
```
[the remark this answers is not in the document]
thejasn: Moved deferred function prior to cleanUp()

pkg/operator/controller/machineconfig/controller.go
```diff
@@ -311,105 +227,52 @@ func (r *MachineConfigReconciler) updateStatus(ctx context.Context) error {
...
+		return false, err
 	}
+	nomc.Status.SetCondition(v1alpha2.DebugEnabled, metav1.ConditionFalse, v1alpha2.ReasonDisabled, "debug configurations disabled")
+	nomc.Status.SetCondition(v1alpha2.DebugReady, metav1.ConditionFalse, v1alpha2.ReasonInProgress, "removing debug configurations in progress")
```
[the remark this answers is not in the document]
thejasn: Moved deferred function prior to cleanUp()

pkg/operator/controller/machineconfig/controller.go
```diff
@@ -80,228 +86,138 @@ func New(mgr ctrl.Manager) *MachineConfigReconciler {
...
 			r.Log.V(1).Info("nodeobservabilitymachineconfig resource not found. Ignoring as it could have been deleted")
 			return ctrl.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
-		return ctrl.Result{RequeueAfter: defaultRequeueTime}, fmt.Errorf("failed to fetch nodeobservabilitymachineconfig: %w", err)
+		return ctrl.Result{}, fmt.Errorf("failed to fetch nodeobservabilitymachineconfig %q: %w", req.NamespacedName, err)
 	}
 
-	r.Log.V(1).Info("nodeobservabilitymachineconfig resource found", "nodeselector", r.CtrlConfig.Spec.NodeSelector)
+	r.Log.V(1).Info("reconciling", "nodeobservabilitymachineconfig", req.NamespacedName)
```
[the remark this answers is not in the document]
alebedev87: I think that we can add the whole `NOB` resource, this will show the whole definition, can be useful. `req` doesn't make sense, it should be there already.

pkg/operator/controller/machineconfig/controller.go
```diff
@@ -80,228 +86,138 @@ func New(mgr ctrl.Manager) *MachineConfigReconciler {
...
 		}
-	}
 
-	defer func() {
-		r.CtrlConfig.Status.UpdateLastReconcileTime()
-		if errUpdate := r.updateStatus(ctx); errUpdate != nil {
-			result = ctrl.Result{RequeueAfter: defaultRequeueTime}
-			err = utilerrors.NewAggregate([]error{err, fmt.Errorf("failed to update status: %w", errUpdate)})
+		if !controllerutil.RemoveFinalizer(nomc, finalizer) {
+			r.Log.V(3).Info("failed to remove finalizer nodeobservabilitymachineconfig", "finalizer", finalizer, "nodeobservabilitymachineconfig.name", req.NamespacedName)
```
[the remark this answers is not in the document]
thejasn: So  `RemoveFinalizer()` doesn't return any error. It returns false if the finalizer wasn't found. Hence just the log. We can add an additional check for `ContainsFinalizer()`  but it just gets redundant.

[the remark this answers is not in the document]
alebedev87: As discussed, both pf the finalizer functions from the controller runtime need the subsequent API updates.

pkg/operator/controller/machineconfig/controller.go
```diff
@@ -80,228 +86,139 @@ func New(mgr ctrl.Manager) *MachineConfigReconciler {
...
+			return ctrl.Result{RequeueAfter: defaultRequeueTime}, nil
 		}
-		r.Log.V(1).Info("removed finalizer from nodeobservabilitymachineconfig resource, cleanup complete")
-		return ctrl.Result{}, nil
 	}
 
-	r.Log.V(1).Info("cleanup is not complete, retry")
-
-	return ctrl.Result{RequeueAfter: defaultRequeueTime}, nil
+	r.Log.V(2).Info("reconciliation completed", "nodeobservabilitymachineconfig", req.NamespacedName)
```
alebedev87, line 178:
```suggestion
	r.Log.V(1).Info("reconciliation completed")
```
We report the start of the reconciliation as V1, let's do the se for the end.

pkg/operator/controller/machineconfig/controller.go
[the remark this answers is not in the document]
thejasn: I've removed this log, adding the entire NOB would just add too much noise for now.

thejasn:
/retest 
CI issues with missing images tags (unrelated to NOB)

thejasn:
/label docs-approved
/label px-approved
Since there is no px or docs impact

PR #100, opened 2022-10-11
nob controller: watch all secondary resources

alebedev87:
- Requeue `NodeObservability` if any of its secondary resources is touched. SCC, service and serviceaccounts were missing.
- SCC update function was changed to disregard the order of some fields: volumes, capabilities (saw the fight with OpenShift defaulting).
- DaemonSet volumesChanged function was changed to disregard `defaultMode` field (saw the fight with OpenShift defaulting).

SCC:
```
$ oc logs deploy/node-observability-operator-controller-manager -c manager | grep -c 'updated securitycontextconstraints'
0

$ oc patch scc node-observability-agent --type='json' -p='[{"op": "replace", "path": "/allowPrivilegedContainer", "value":false}]'
securitycontextconstraints.security.openshift.io/node-observability-agent patched

$ oc logs deploy/node-observability-operator-controller-manager -c manager | grep -c 'updated securitycontextconstraints'
1
```

Service:
```
$ oc get svc node-observability-agent
NAME                       TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)    AGE
node-observability-agent   ClusterIP   None         <none>        8443/TCP   162m

$ oc logs deploy/node-observability-operator-controller-manager -c manager | grep -c 'updated service'
0

$ oc patch svc node-observability-agent --type='json' -p='[{"op": "replace", "path": "/spec/ports/0/port", "value":9999}]'
service/node-observability-agent patched

$ oc logs deploy/node-observability-operator-controller-manager -c manager | grep -c 'updated service'
1
```

ServiceAccount:
```
$ oc get sa node-observability-agent
NAME                       SECRETS   AGE
node-observability-agent   1         156m

$ oc logs deploy/node-observability-operator-controller-manager -c manager | grep -c 'created serviceaccount'
0

$ oc delete sa node-observability-agent
serviceaccount "node-observability-agent" deleted

$ oc logs deploy/node-observability-operator-controller-manager -c manager | grep -c 'created serviceaccount'
1
```

PR #102, opened 2022-10-12
nob controller: use emptyDir for profiling data

alebedev87:
This PR uses `emptyDir` volume for the profiling data generated by the agent. This will match the data's lifetime to POD's as opposed to the container's.

Other changes done along the way:
- `volumesChanged` function uses current volumes as the starting point while it has to use the expected ones. This resulted into duplicate volumes while updating the agent.
- `volumeMountsChanged` was aligned with `volumesChanged`.


alebedev87:
Needs to wait till https://github.com/openshift/node-observability-operator/pull/100 is merged to add the code to ignore `defaultMode` in `emptyDir` volumes.

/hold

pkg/operator/controller/nodeobservability/daemonset.go
```diff
@@ -248,6 +254,12 @@ func (r *NodeObservabilityReconciler) desiredDaemonSet(nodeObs *v1alpha2.NodeObs
...
 							},
 						},
+						{
+							Name: dataVolumeName,
+							VolumeSource: corev1.VolumeSource{
+								EmptyDir: &corev1.EmptyDirVolumeSource{},
+							},
+						},
 					},
 					NodeSelector: nodeObs.Spec.NodeSelector,
```
thejasn, line 264: Can you add termination grace period to the daemonset pod spec? (related to https://github.com/openshift/node-observability-agent/pull/38)

alebedev87: Yes, sure will do: once the agent PR is merged, I'll bump the new version of the agent with the needed grace period.

pkg/operator/controller/nodeobservability/daemonset_test.go
```diff
@@ -144,6 +147,7 @@ func TestEnsureDaemonset(t *testing.T) {
 				withHostPathVolume(socketName, socketPath, corev1.HostPathSocket).
 				withConfigMapVolume(kbltCAName, nodeObsInstanceName).
 				withSecretVolume(certsName, secretName).
+				withEmptyDirVolume("profiledata").
```
ayesha54, line 150: shall we use "profiledata" and "/run/node-observability" as variables or let them be the hardcoded ones in the test files?

alebedev87: I prefer hard coded in the tests to double check that I use the right thing in the actual code. Because if there is a mistake in the constant used in the code and the same one is used in the tests - both will be wrong.

pkg/operator/controller/nodeobservability/daemonset.go
thejasn: So since #110 is opened, will you add the change in that or this PR?

alebedev87: It's already [there](https://github.com/openshift/node-observability-operator/blob/2e77565788be53849501e5143a87592d44f02c12/pkg/operator/controller/nodeobservability/daemonset.go#L141-L143).

alebedev87:
Doc PR: https://github.com/openshift/openshift-docs/pull/52213

alebedev87:
/label px-approved

As per Chris Fields: updated docs are enough for this release of NOB operator.

PR #104, opened 2022-10-13
CFE-628: Avoid unnecessary reboot of the worker nodes

sairameshv:
If the UNIX Socket profiling is already enabled in the CRI-O, it is not required to create a nodeObservabilityMachineConfig object that is responsible for the unwanted reboot of the worker mcp

Signed-off-by: Sai Ramesh Vanka <svanka@redhat.com>


pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -19,6 +19,11 @@ package nodeobservabilitycontroller
 import (
 	"context"
 	"fmt"
+	"strings"
```
thejasn, line 22: Let's preserve the forrmatting order. It goes:

1. [k8s.io/](http://k8s.io/)
2. [github.com/openshift/api](http://github.com/openshift/api)
3. 3rd-party libraries (github, sigs, etc)
4. our own module.

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -313,7 +322,34 @@ func (r *NodeObservabilityReconciler) ensureNodeObservabilityDeleted(ctx context
 
 // machineConfigChangeRequested returns true, when a given NodeObservabilityType needs
 // machine config change. Only CrioKubeletNodeObservabilityType requires a MC change, false otherwise
-func machineConfigChangeRequested(nodeObs *operatorv1alpha2.NodeObservability) bool {
+func (r *NodeObservabilityReconciler) machineConfigChangeRequested(ctx context.Context, nodeObs *operatorv1alpha2.NodeObservability) bool {
+	// avoid the creation of a new NOMC if the CRIO profiling is already enabled in the rendered MC
+	mcpName := types.NamespacedName{Name: workerMCP}
+	mcp := &mcv1.MachineConfigPool{}
+	err := r.ClusterWideClient.Get(ctx, mcpName, mcp)
```
thejasn, line 329: `ClusterWideClient` is mainly only included for accessing resources across namespaces. Cluster scoped resources we should be fine accessing using the just `Client`.

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -313,7 +322,34 @@ func (r *NodeObservabilityReconciler) ensureNodeObservabilityDeleted(ctx context
...
+	if err != nil {
+		r.Log.Error(err, "failed to get the machineconfigpool instance "+mcpName.Name)
+		return nodeObs.Spec.Type == operatorv1alpha2.CrioKubeletNodeObservabilityType
+	}
```
thejasn, line 333:
```suggestion
        enabled := nodeObs.Spec.Type == operatorv1alpha2.CrioKubeletNodeObservabilityType
        if !enabled {
               r.Log.Info("unsupported profiling type provided","nodeobservability.name", nodeObs.Name)
               return enabled
        }
        
	mcpName := types.NamespacedName{Name: workerMCP}
	mcp := &mcv1.MachineConfigPool{}
	err := r.ClusterWideClient.Get(ctx, mcpName, mcp)
	if err != nil {
		r.Log.Error(err, "failed to get the machineconfigpool instance "+mcpName.Name)
		return enabled
	}
```

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -313,7 +322,34 @@ func (r *NodeObservabilityReconciler) ensureNodeObservabilityDeleted(ctx context
...
+	// getting the rendered mc name from the obtained machine config pool resource
+	mcName := types.NamespacedName{Name: mcp.Spec.Configuration.Name}
+	renderedMC := &mcv1.MachineConfig{}
+	err = r.ClusterWideClient.Get(ctx, mcName, renderedMC)
+	if err != nil {
+		r.Log.Error(err, "failed to get the machineconfig instance "+mcName.Name)
+		return nodeObs.Spec.Type == operatorv1alpha2.CrioKubeletNodeObservabilityType
```
thejasn, line 340:
```suggestion
		return enabled
```

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -313,7 +322,34 @@ func (r *NodeObservabilityReconciler) ensureNodeObservabilityDeleted(ctx context
...
+	}
+	bData, err := yaml.Marshal(renderedMC.Spec.Config)
+	if err != nil {
+		r.Log.Error(err, "failed to get CRIO profiling info from the rendered machineconfig "+renderedMC.Name)
+		return nodeObs.Spec.Type == operatorv1alpha2.CrioKubeletNodeObservabilityType
```
thejasn, line 345:
```suggestion
		return enabled
```

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -313,7 +322,34 @@ func (r *NodeObservabilityReconciler) ensureNodeObservabilityDeleted(ctx context
...
+	}
+	if bData != nil {
+		if strings.Contains(string(bData), crioProfilingInfo) {
+			// profiling has already been enabled on the CRIO and hence not creating a new NOMC
+			return false
```
thejasn, line 350:
```suggestion
			enabled = false
```

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -313,7 +322,34 @@ func (r *NodeObservabilityReconciler) ensureNodeObservabilityDeleted(ctx context
...
+		}
+	}
 	return nodeObs.Spec.Type == operatorv1alpha2.CrioKubeletNodeObservabilityType
```
thejasn, line 353:
```suggestion
	return enabled
```

thejasn (changes requested):

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -313,7 +322,34 @@ func (r *NodeObservabilityReconciler) ensureNodeObservabilityDeleted(ctx context
...
+	}
+	// getting the rendered mc name from the obtained machine config pool resource
+	mcName := types.NamespacedName{Name: mcp.Spec.Configuration.Name}
+	renderedMC := &mcv1.MachineConfig{}
+	err = r.ClusterWideClient.Get(ctx, mcName, renderedMC)
+	if err != nil {
+		r.Log.Error(err, "failed to get the machineconfig instance "+mcName.Name)
+		return nodeObs.Spec.Type == operatorv1alpha2.CrioKubeletNodeObservabilityType
+	}
+	bData, err := yaml.Marshal(renderedMC.Spec.Config)
```
thejasn, line 342: nit: `machineconfig` controller uses `json` to marshal/unmarshal, we could keep both in sync.

sairameshv: I figured out during my testing that this won't work as the Config has a nested data.
Anyway, there is a Marshal() method for the `Spec.Config` that's working (verified)

pkg/operator/controller/nodeobservability/controller.go
sairameshv: Addressed

pkg/operator/controller/nodeobservability/controller.go
sairameshv: Addressed

pkg/operator/controller/nodeobservability/controller.go
sairameshv: Addressed

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -313,8 +321,39 @@ func (r *NodeObservabilityReconciler) ensureNodeObservabilityDeleted(ctx context
...
+	if err != nil {
+		r.Log.Error(err, "failed to get the machineconfigpool instance "+mcpName.Name)
+		return isChangeRequested
+	}
+	// getting the rendered mc name from the obtained machine config pool resource
+	mcName := types.NamespacedName{Name: mcp.Spec.Configuration.Name}
+	renderedMC := &mcv1.MachineConfig{}
+	err = r.Client.Get(ctx, mcName, renderedMC)
+	if err != nil {
+		r.Log.Error(err, "failed to get the machineconfig instance "+mcName.Name)
```
alebedev87, line 342:
```suggestion
		r.Log.Error(err, "failed to get the machineconfig instance", "mc.name", mcName.Name)
```

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -313,8 +321,39 @@ func (r *NodeObservabilityReconciler) ensureNodeObservabilityDeleted(ctx context
...
+		return isChangeRequested
+	}
+	bData, err := renderedMC.Spec.Config.Marshal()
+	if err != nil {
+		r.Log.Error(err, "failed to get CRIO profiling info from the rendered machineconfig "+renderedMC.Name)
```
alebedev87, line 347:
```suggestion
		r.Log.Error(err, "failed to get CRIO profiling info from the rendered machineconfig", "renderedmc.name", renderedMC.Name)
```

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -313,8 +321,39 @@ func (r *NodeObservabilityReconciler) ensureNodeObservabilityDeleted(ctx context
...
+		return isChangeRequested
+	}
+	if bData != nil {
+		if strings.Contains(string(bData), crioProfilingInfo) {
+			// profiling has already been enabled on the CRIO and hence not creating a new NOMC
```
alebedev87, line 352: Would be good to log the decision to skip the machine configuration.

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -313,8 +321,39 @@ func (r *NodeObservabilityReconciler) ensureNodeObservabilityDeleted(ctx context
...
+	isChangeRequested := nodeObs.Spec.Type == operatorv1alpha2.CrioKubeletNodeObservabilityType
+	if !isChangeRequested {
+		return isChangeRequested
+	}
+	// avoid the creation of a new NOMC if the CRIO profiling is already enabled in the rendered MC
+	mcpName := types.NamespacedName{Name: workerMCP}
+	mcp := &mcv1.MachineConfigPool{}
+	err := r.Client.Get(ctx, mcpName, mcp)
+	if err != nil {
+		r.Log.Error(err, "failed to get the machineconfigpool instance "+mcpName.Name)
```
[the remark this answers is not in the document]
sairameshv: We ended up in an error situation and hence no data of the `mcp.Name` present. Moreover, both `mcp.Name`, `mcpName.Name` are one and the same.

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -19,25 +19,33 @@ package nodeobservabilitycontroller
...
 )
 
 const (
 	finalizer = "NodeObservability"
 	// the name of the NodeObservability resource which will be reconciled
 	nodeObsCRName = "cluster"
+	// workerMCP refers to the name of the worker machine config pool instance
+	workerMCP = "worker"
+	// crioProfilingInfo refers to the env data that enables the profiling in the CRI-O
+	crioProfilingInfo = "ENABLE_PROFILE_UNIX_SOCKET=true"
```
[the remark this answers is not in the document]
sairameshv: Addressed, Defined a new constant there as the existing constant contains extra info which is not helpful while finding the substring in the rendered machine config data.

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -313,8 +321,39 @@ func (r *NodeObservabilityReconciler) ensureNodeObservabilityDeleted(ctx context
...
+func (r *NodeObservabilityReconciler) machineConfigChangeRequested(ctx context.Context, nodeObs *operatorv1alpha2.NodeObservability) bool {
+	isChangeRequested := nodeObs.Spec.Type == operatorv1alpha2.CrioKubeletNodeObservabilityType
+	if !isChangeRequested {
+		return isChangeRequested
+	}
+	// avoid the creation of a new NOMC if the CRIO profiling is already enabled in the rendered MC
+	mcpName := types.NamespacedName{Name: workerMCP}
+	mcp := &mcv1.MachineConfigPool{}
+	err := r.Client.Get(ctx, mcpName, mcp)
+	if err != nil {
```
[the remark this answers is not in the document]
sairameshv: Addressed

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -313,8 +321,39 @@ func (r *NodeObservabilityReconciler) ensureNodeObservabilityDeleted(ctx context
...
+		r.Log.Error(err, "failed to get the machineconfigpool instance "+mcpName.Name)
+		return isChangeRequested
+	}
+	// getting the rendered mc name from the obtained machine config pool resource
+	mcName := types.NamespacedName{Name: mcp.Spec.Configuration.Name}
+	renderedMC := &mcv1.MachineConfig{}
+	err = r.Client.Get(ctx, mcName, renderedMC)
+	if err != nil {
```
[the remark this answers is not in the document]
sairameshv: Addressed

pkg/operator/controller/nodeobservability/controller.go
sairameshv: Addressed

pkg/operator/controller/nodeobservability/controller.go
sairameshv: Addressed

pkg/operator/controller/nodeobservability/controller.go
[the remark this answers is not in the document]
sairameshv: Addressed

pkg/operator/controller/machineconfig/constants.go
```diff
@@ -41,6 +41,10 @@ const (
 	CrioUnixSocketConfData = `[Service]
 Environment="ENABLE_PROFILE_UNIX_SOCKET=true"`
 
+	// CrioUnixSocketEnvString refers to the environment variable info
+	// string that helps in finding if the profiling is enabled by default
+	CrioUnixSocketEnvString = "ENABLE_PROFILE_UNIX_SOCKET=true"
```
alebedev87, line 46: This variable can be used in the conf file data (variable just above). Example: https://go.dev/play/p/2WYvJqUgekV

sairameshv: The only reason I didn't make this change is to maintain it as a constant instead of a variable. WDYT?

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -335,8 +339,41 @@ func (r *NodeObservabilityReconciler) ensureNodeObservabilityDeleted(ctx context
...
 // machineConfigChangeRequested returns true, when a given NodeObservabilityType needs
 // machine config change. Only CrioKubeletNodeObservabilityType requires a MC change, false otherwise
-func machineConfigChangeRequested(nodeObs *operatorv1alpha2.NodeObservability) bool {
-	return nodeObs.Spec.Type == operatorv1alpha2.CrioKubeletNodeObservabilityType
+func (r *NodeObservabilityReconciler) machineConfigChangeRequested(ctx context.Context, nodeObs *operatorv1alpha2.NodeObservability) bool {
+	isChangeRequested := nodeObs.Spec.Type == operatorv1alpha2.CrioKubeletNodeObservabilityType
+	if !isChangeRequested {
+		if nodeObs.Spec.Type != "" {
+			r.Log.Error(nil, "unsupoorted node observability type found", "type", nodeObs.Spec.Type)
+		}
```
alebedev87, line 347: This is unnecessary as there cannot be any other type for the moment: [openapi validation](https://github.com/openshift/node-observability-operator/blob/319c83f0300b327641a3f39d405a952b67f49e03/api/v1alpha2/nodeobservability_types.go#L23-L24).

sairameshv: Addressed

pkg/operator/controller/machineconfig/constants.go
alebedev87: Once both variables are tight together, we make a link between 2 controllers and this will help us to not forget to update both when the time will come to think about scheduling on other node types (not only workers). We can sacrifice 1 compile time substitution for that.

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -335,8 +339,38 @@ func (r *NodeObservabilityReconciler) ensureNodeObservabilityDeleted(ctx context
...
+	// getting the rendered mc name from the obtained machine config pool resource
+	mcName := types.NamespacedName{Name: mcp.Spec.Configuration.Name}
+	renderedMC := &mcv1.MachineConfig{}
+	if err := r.Client.Get(ctx, mcName, renderedMC); err != nil {
+		r.Log.Error(err, "failed to get the machineconfig instance ", "mc.name", mcName.Name)
+		return isChangeRequested
+	}
+	bData, err := renderedMC.Spec.Config.Marshal()
+	if err != nil {
+		r.Log.Error(err, "failed to get CRIO profiling info from the rendered machineconfig ", "renderedMc.name", renderedMC.Name)
```
alebedev87, line 363:
```suggestion
		r.Log.Error(err, "failed to marshal rendered machineconfig ", "mc.name", renderedMC.Name)
```

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -202,7 +206,7 @@ func (r *NodeObservabilityReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	// if machine config change is not requested, we can mark it as ready
 	var nomcReady bool = true
```
alebedev87, line 208:
```suggestion
	var mcReady bool = true
```

Would be good to modify the log message on the line 218 from `NodeObservabilityMachineConfig` to something like `MachineConfig` or `MachineConfiguration`. As now the machine can be ready without `NOBMC` CR being involved.

sairameshv:
/label px-approved
(As docs-approved label is received)

PR #107, opened 2022-10-19
CFE-600: Bump the openshift/api, k8s and the dependent packages

sairameshv:
The following are the new versions of the bumped packages
```
1. openshift/api v0.0.0-20221013123531-622889ac07cf
2. k8s.io/api v0.25.0
3. k8s.io/apimachinery v0.25.0
4. k8s.io/client-go v0.25.0
5. k8s.io/utils v0.0.0-20220728103510-ee6ede2d64ed
6. sigs.k8s.io/controller-runtime v0.13.0
7. golangci-lint v1.50.0

```
As the "`k8s.io/apimachinery/pkg/util/clock`" is deprecated, it has been replaced by the "`k8s.io/utils/clock`" package
Similarly, the deprecated `io/ioutil` is replaced by the `io`, `os` packages wherever required.
Fixed minor issues caused by the updated golangci-lint package.

Signed-off-by: Sai Ramesh Vanka <svanka@redhat.com>


sairameshv:
CI tests are failing as the current ci env doesn't have the go version >= 1.19.
Raised a PR for the same.
https://github.com/openshift/release/pull/33237

sairameshv:
/retest
Triggering the CI tests as the ci env is bumped to have go 1.19 version

go.mod
```diff
@@ -1,36 +1,36 @@
 module github.com/openshift/node-observability-operator
 
-go 1.17
+go 1.18
```
alebedev87, line 3: Looking at the logs of the CI I see that `golang-1.18` image is still used for the builds,seems like `build_root` field from the CI config was exactly for this. However the operator image is built from `Dockerfile.rhel8` image ([link](https://github.com/openshift/release/blob/918f7fefade600a73f960056167d772f4ca35eb8/ci-operator/config/openshift/node-observability-operator/openshift-node-observability-operator-main.yaml#L23-L30)) which uses golang 1.19.
We should avoid this discrepancy. Since `build_root` doesn't have 1.19 yet we can try to instruct CI to look for the build image in the repository. For instance, similar to the way it's done in extdns operator: [repo](https://github.com/openshift/external-dns-operator/blob/main/.ci-operator.yaml), [CI config](https://github.com/openshift/release/blob/918f7fefade600a73f960056167d772f4ca35eb8/ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-main.yaml#L18-L19).

go.mod
```diff
@@ -1,36 +1,36 @@
...
+	github.com/openshift/build-machinery-go v0.0.0-20220913142420-e25cf57ea46d
 	github.com/openshift/machine-config-operator v0.0.1-0.20220201192635-14a1ca2cb91f
-	go.uber.org/zap v1.19.1
-	k8s.io/api v0.23.4
-	k8s.io/apimachinery v0.23.4
-	k8s.io/client-go v0.23.4
-	k8s.io/utils v0.0.0-20211116205334-6203023598ed
-	sigs.k8s.io/controller-runtime v0.11.1
+	go.uber.org/zap v1.21.0
+	k8s.io/api v0.25.0
```
alebedev87, line 17: The latest available is `v0.25.3`.

go.mod
sairameshv: Related PRs: 
1. https://github.com/openshift/node-observability-operator/pull/108
2. https://github.com/openshift/release/pull/33281

go.mod
sairameshv: Addressed

go.mod
```diff
@@ -1,36 +1,36 @@
 module github.com/openshift/node-observability-operator
 
-go 1.17
+go 1.19
 
 require (
 	github.com/coreos/ignition/v2 v2.13.0
-	github.com/go-logr/logr v1.2.2
+	github.com/go-logr/logr v1.2.3
 	github.com/golangci/golangci-lint v1.42.1
```
alebedev87, line 8: This needs to be updated to the latest version, to prevent the OOM we saw in release repository.

sairameshv: Bumping to the latest version 1.50.0

alebedev87:
/label docs-approved
/label px-approved

No doc updates needed.

PR #109, opened 2022-10-24
nob run controller: brushing

alebedev87:


pkg/operator/controller/nodeobservability/daemonset.go
```diff
@@ -14,6 +14,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	v1alpha2 "github.com/openshift/node-observability-operator/api/v1alpha2"
+	opctrl "github.com/openshift/node-observability-operator/pkg/operator/controller"
```
ayesha54, line 17: I think we can keep the same name for the controller (or any other package) wherever we are importing it, just for easy readability.

pkg/operator/controller/nodeobservabilityrun/controller.go
```diff
@@ -80,114 +80,105 @@ func (r *NodeObservabilityRunReconciler) Reconcile(ctx context.Context, req ctrl
...
 
 	// one cycle takes cca 30s
-	return ctrl.Result{RequeueAfter: time.Second * 30}, err
+	return ctrl.Result{RequeueAfter: time.Second * 30}, nil
 }
 
-// adoptResource sets OwnerReference to point to NodeObservability from .spec.ref.name
-// returns true if already adopted, false otherwise
+// adoptResource sets OwnerReference to point to NodeObservability from .spec.nodeobservabilityref.name.
+// returns true if was adopted by this function, false otherwise.
```
ayesha54, line 152: // returns true if it was adopted by this function, false otherwise.

pkg/operator/controller/nodeobservability/daemonset.go
alebedev87: From what I can see it's `opctrl` everywhere:
```
$ agi 'github.com/openshift/node-observability-operator/pkg/operator/controller\"'
pkg/operator/controller/nodeobservability/daemonset.go
17:	opctrl "github.com/openshift/node-observability-operator/pkg/operator/controller"

pkg/operator/controller/nodeobservability/service.go
20:	opctrl "github.com/openshift/node-observability-operator/pkg/operator/controller"

pkg/operator/controller/nodeobservabilityrun/controller.go
47:	opctrl "github.com/openshift/node-observability-operator/pkg/operator/controller"
```

pkg/operator/controller/nodeobservabilityrun/controller.go
alebedev87: Got it, I missed `it`, I'm wondering whether the way it's now is a invalid English.
Will add `it` if some other comments will pop up.

pkg/operator/controller/nodeobservability/daemonset.go
ayesha54: Yes, because I also used opctrl in my PR after seeing this.

.golangci.yaml
```diff
@@ -79,10 +78,8 @@ linters:
     - ineffassign
     - misspell
     - staticcheck
-    - structcheck
     - typecheck
     - unused
-    - varcheck
```
ayesha54, line 85: That's a much needed change for the clean make verify

pkg/operator/controller/names.go
```diff
@@ -1,5 +1,23 @@
...
+limitations under the License.
+*/
+
 package nodeobservabilitycontroller
 
 const (
-	AgentName = "node-observability-agent"
+	// AgentServiceName is the name of the service backed up by the agent's daemonset.
+	AgentServiceName = "node-observability-agent"
+	// nolint
```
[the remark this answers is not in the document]
alebedev87: because `golangci-lint` doesn't like when `secret` is a part of the variable name, which is non sense. To not disable the check completely - I excluded only this variable.
Will change the comment to describe this, thanks!

pkg/operator/controller/nodeobservabilityrun/controller.go
```diff
@@ -80,114 +80,105 @@ func (r *NodeObservabilityRunReconciler) Reconcile(ctx context.Context, req ctrl
...
 	}
 
-	msg = "Profiling query initiated"
-	instance.Status.SetCondition(nodeobservabilityv1alpha2.DebugFinished, metav1.ConditionFalse, nodeobservabilityv1alpha2.ReasonInProgress, msg)
+	r.Log.V(1).Info("profiling query initiated")
+	instance.Status.SetCondition(nodeobservabilityv1alpha2.DebugFinished, metav1.ConditionFalse, nodeobservabilityv1alpha2.ReasonInProgress, "Profiling query initiated")
 
 	// one cycle takes cca 30s
-	return ctrl.Result{RequeueAfter: time.Second * 30}, err
+	return ctrl.Result{RequeueAfter: time.Second * 30}, nil
```
ayesha54, line 148: why nil and remove the err?

sherine-k (approved):
/approve

alebedev87:
/label px-approved

As per Chris Fields, the docs are sufficient for this release.

PR #111, opened 2022-10-26
CFE-511: adding kubelet-ca configmap controller

ayesha54:
Node Observability Agent uses TLS to talk to kubelet. Currently, the kubelet CA is copied from [the openshift config namespace](https://github.com/openshift/node-observability-operator/blob/main/pkg/operator/controller/nodeobservability/configmap.go#L22-L25) at the time of [the creation of the agent's daemnset](https://github.com/openshift/node-observability-operator/blob/main/pkg/operator/controller/nodeobservability/daemonset.go#L41).
This prevents the agent from following up with the CA rotations because:

target (in the operator's namespace) configmap is created and never updated
coupling of the daemonset lifecycle with the target config map
To address the before mentioned points a dedicated controller has to be added to the operator. This controller would have to copy the configmap from the source namespace into the target one.
Note: the agent's daemonset would have to be able to reload the contents of the configmap into its TLS connection settings (fswatch or annotation with the hash).
Example from another operators: [ca-configmap controller](https://github.com/openshift/external-dns-operator/tree/main/pkg/operator/controller/ca-configmap)

Acceptance criteria:

Unit tests covering the new controller and changes in the nodeobservability controller
E2E is working

https://issues.redhat.com/browse/CFE-511


bundle/manifests/node-observability-operator.clusterserviceversion.yaml
```diff
@@ -392,7 +391,7 @@ spec:
                       fieldPath: metadata.namespace
                 - name: RELATED_IMAGE_AGENT
                   value: quay.io/node-observability-operator/node-observability-agent@sha256:ac858cd54e895a61de6fb537d79a3f1245ffb4a1f12c81bade7d28168bd5951f
-                image: quay.io/openshift/node-observability-operator:v0.0.1
+                image: quay.io/akaleem/node-observability-operator:latest
```
alebedev87, line 394: This should be reverted to the initial value.

pkg/operator/controller/names.go
```diff
@@ -1,5 +1,34 @@
...
+limitations under the License.
+*/
+
 package nodeobservabilitycontroller
 
+import "k8s.io/apimachinery/pkg/types"
+
 const (
-	AgentName = "node-observability-agent"
+	// nodeObservabilityBaseName  is the base name for any NodeObservability resource.
```
alebedev87, line 22: This doesn't relate to anything.

pkg/operator/operator.go
```diff
@@ -84,28 +86,49 @@ func New(cliCfg *rest.Config, opCfg *operatorconfig.Config) (*Operator, error) {
...
-		return nil, fmt.Errorf("failed to create client: %w", err)
+		return nil, fmt.Errorf("failed to create controller-runtime cluster: %w", err)
+	}
+
+	if err := mgr.Add(cluster); err != nil {
+		return nil, fmt.Errorf("failed to add controller-runtime cluster to manager: %w", err)
 	}
+	// Create and register the CA config map controller with the operator manager.
+	if _, err := caconfigmapcontroller.New(mgr, caconfigmapcontroller.Config{
+		SourceNamespace: "openshift-config-managed",
```
alebedev87, line 115: This is better to be a constant in this file.

pkg/operator/operator.go
```diff
@@ -84,28 +86,49 @@ func New(cliCfg *rest.Config, opCfg *operatorconfig.Config) (*Operator, error) {
...
+		TargetNamespace: opCfg.OperatorNamespace,
+		CAConfigMapName: "kubelet-serving-ca",
```
alebedev87, line 117: This should come from `names.go`.

pkg/operator/controller/utils/test/vars.go
```diff
@@ -23,6 +23,7 @@ import (
 const (
 	TestNamespace     = ""
 	OperatorNamespace = "node-observability-operator"
+	SourceNamespace   = "openshift-config-managed"
```
alebedev87, line 26:
```suggestion
	KubeletCASourceNamespace   = "openshift-config-managed"
```

pkg/operator/controller/nodeobservability/daemonset_test.go
```diff
@@ -33,11 +33,14 @@ import (
...
 	operatorv1alpha2 "github.com/openshift/node-observability-operator/api/v1alpha2"
-	"github.com/openshift/node-observability-operator/pkg/operator/controller/test"
+	"github.com/openshift/node-observability-operator/pkg/operator/controller/utils/test"
 )
 
 const (
-	nodeObsInstanceName = "nodeobservability-sample"
+	nodeObsInstanceName         = "nodeobservability-sample"
+	kubeletCAConfigMapName      = "kubelet-serving-ca"
+	srcKbltCAConfigMapName      = "kubelet-serving-ca"
```
alebedev87, line 42: Why 2 vars? The configmap doesn't change in the source or target namespaces.

pkg/operator/controller/names.go
```diff
@@ -1,5 +1,34 @@
...
+	AgentName      = "node-observability-agent"
 )
+
+// NodeObsKubeletCAConfigMapName returns the namespaced name of the destination (operand) kubelet CA configmap
+func NodeObsKubeletCAConfigMapName(operandNamespace string) types.NamespacedName {
+	return types.NamespacedName{
+		Namespace: operandNamespace,
+		Name:      kubeletCM,
+	}
+}
```
alebedev87, line 34: I don't quite see the helpfulness of this function, I think it makes more sense to expose kubelet configMap name and use it with whatever namespace is needed.

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -195,8 +197,21 @@ func (r *NodeObservabilityReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	configMapExists, configMap, err := r.currentNodeObsKubeletCAConfigMap(ctx, configMapNsName)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to get the target CA configmap %q: %w", configMapNsName, err)
+	}
+	if !configMapExists {
+		// kubelet CA configmap was not synced yet or doesn't exist at all,
+		// either way: no need to requeue immediately polluting the logs.
+		return reconcile.Result{RequeueAfter: RequeuePeriod(5)}, fmt.Errorf("target CA configmap %q not found", configMapNsName)
+	}
+	kubeletCAConfigMap = configMap
```
alebedev87, line 212:
```suggestion
	configMapNsName := types.NamespacedName{Namespace:r.Namespace, name: opctrl.KubeletCAConfigMapName}
	configMapExists, kubeletCAConfigMap, err := r.currentNodeObsKubeletCAConfigMap(ctx, configMapNsName)
	if err != nil {
		return reconcile.Result{}, fmt.Errorf("failed to get the target CA configmap %q: %w", configMapNsName, err)
	}
	if !configMapExists {
		// kubelet CA configmap was not synced yet or doesn't exist at all,
		// either way: no need to requeue immediately polluting the logs.
		return reconcile.Result{RequeueAfter: defaultRequeuePeriod}, fmt.Errorf("target CA configmap %q not found", configMapNsName)
	}
```

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -381,3 +396,21 @@ func isClusterNodeObservability(ctx context.Context, nodeObs *operatorv1alpha2.N
 
 	return fmt.Errorf("a single NodeObservability with name 'cluster' is authorized. Resource %s will be ignored", nodeObs.Name)
 }
+
+// RequeuePeriod returns requeue period as time.Duration.
+func RequeuePeriod(RequeuePeriodSeconds int) time.Duration {
+	return time.Duration(RequeuePeriodSeconds) * time.Second
+}
```
alebedev87, line 403: This can be replaced by 1 constant I mentioned in the suggestion above: `defaultRequeuePeriod`.

pkg/operator/controller/nodeobservability/daemonset.go
```diff
@@ -34,31 +34,29 @@ const (
...
+	var (
+		err     error
+		desired *appsv1.DaemonSet
+	)
+	if kubeletCAConfigMap != nil {
+		desired = r.desiredDaemonSet(nodeObs, sa, ns, kubeletCAConfigMap.Name)
+		if err := controllerutil.SetControllerReference(nodeObs, desired, r.Scheme); err != nil {
+			return nil, fmt.Errorf("failed to set the controller reference for daemonset: %w", err)
+		}
 	}
```
alebedev87, line 47:
```suggestion
	desired = r.desiredDaemonSet(nodeObs, sa, ns, kubeletCAConfigMap.Name)
	if err := controllerutil.SetControllerReference(nodeObs, desired, r.Scheme); err != nil {
		return nil, fmt.Errorf("failed to set the controller reference for daemonset: %w", err)
	}
```
We never pass empty kubelet configmap, it doesn't make sense to because we need it for the daemonset, w/o it the daemonset cannot be created.

pkg/operator/controller/names.go
ayesha54: that's what this function is doing, and it's also used by the ca-configmap controller.

pkg/operator/controller/names.go
```diff
@@ -1,5 +1,34 @@
...
+	SrcKbltCAConfigMapNamespace = "openshift-config-managed"
+	ControllerName              = "node_observability_controller"
+	AgentName                   = "node-observability-agent"
+	KbltCAConfigMapName         = "kubelet-serving-ca"
 )
+
+// NamespacedKubeletCAConfigMapName returns the namespaced name of the kubelet CA configmap with the provided namespace.
+func NamespacedKubeletCAConfigMapName(operandNamespace string) types.NamespacedName {
+	return types.NamespacedName{
+		Namespace: operandNamespace,
```
alebedev87, line 31:
```suggestion
func NamespacedKubeletCAConfigMapName(namespace string) types.NamespacedName {
	return types.NamespacedName{
		Namespace: namespace,
```

pkg/operator/controller/names.go
```diff
@@ -1,5 +1,34 @@
...
+limitations under the License.
+*/
+
 package nodeobservabilitycontroller
 
+import "k8s.io/apimachinery/pkg/types"
+
 const (
-	AgentName = "node-observability-agent"
+	SrcKbltCAConfigMapNamespace = "openshift-config-managed"
```
alebedev87, line 22:
```suggestion
	SourceKubeletCAConfigMapNamespace = "openshift-config-managed"
```
I liked how you called `NamespacedKubeletCAConfigMapName` - everything is very clear. Let's do the same for this constant.

pkg/operator/controller/names.go
```diff
@@ -1,5 +1,34 @@
...
+	ControllerName              = "node_observability_controller"
+	AgentName                   = "node-observability-agent"
+	KbltCAConfigMapName         = "kubelet-serving-ca"
```
alebedev87, line 25:
```suggestion
	KubeletCAConfigMapName         = "kubelet-serving-ca"
```
Same as for the source namespace constant.

pkg/operator/controller/names.go
```diff
@@ -1,5 +1,34 @@
...
+*/
+
 package nodeobservabilitycontroller
 
+import "k8s.io/apimachinery/pkg/types"
+
 const (
-	AgentName = "node-observability-agent"
+	SrcKbltCAConfigMapNamespace = "openshift-config-managed"
+	ControllerName              = "node_observability_controller"
```
alebedev87, line 23: This one needs to be removed as it's not used anywhere.

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -195,8 +198,21 @@ func (r *NodeObservabilityReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	configMapExists, configMap, err := r.currentNodeObsKubeletCAConfigMap(ctx, configMapNsName)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to get the target CA configmap %q: %w", configMapNsName, err)
+	}
+	if !configMapExists {
+		// kubelet CA configmap was not synced yet or doesn't exist at all,
+		// either way: no need to requeue immediately polluting the logs.
+		return reconcile.Result{RequeueAfter: defaultRequeuePeriod}, fmt.Errorf("target CA configmap %q not found", configMapNsName)
+	}
+	kubeletCAConfigMap = configMap
```
alebedev87, line 213:
```suggestion
	configMapNsName := opctrl.NamespacedKubeletCAConfigMapName(r.Namespace)
	configMapExists, kubeletCAConfigMap, err := r.currentNodeObsKubeletCAConfigMap(ctx, configMapNsName)
	if err != nil {
		return reconcile.Result{}, fmt.Errorf("failed to get the target CA configmap %q: %w", configMapNsName, err)
	}
	if !configMapExists {
		// kubelet CA configmap was not synced yet or doesn't exist at all,
		// either way: no need to requeue immediately polluting the logs.
		return reconcile.Result{RequeueAfter: defaultRequeuePeriod}, fmt.Errorf("target CA configmap %q not found", configMapNsName)
	}
```

pkg/operator/controller/utils/test/vars.go
```diff
@@ -21,8 +21,9 @@ import (
 )
 
 const (
-	TestNamespace     = ""
-	OperatorNamespace = "node-observability-operator"
+	TestNamespace            = ""
+	OperatorNamespace        = "node-observability-operator"
+	KubeletCASourceNamespace = "openshift-config-managed"
```
sherine-k, line 26: Do you prefer to redefine the namespaces here, as opposed to reusing from names.go?

pkg/operator/operator.go
```diff
@@ -84,28 +86,49 @@ func New(cliCfg *rest.Config, opCfg *operatorconfig.Config) (*Operator, error) {
...
+
+	if err := mgr.Add(cluster); err != nil {
+		return nil, fmt.Errorf("failed to add controller-runtime cluster to manager: %w", err)
 	}
+	// Create and register the CA config map controller with the operator manager.
+	if _, err := caconfigmapcontroller.New(mgr, caconfigmapcontroller.Config{
+		SourceNamespace: opctrl.SourceKubeletCAConfigMapNamespace,
+		TargetNamespace: opCfg.OperatorNamespace,
+		CAConfigMapName: opctrl.KubeletCAConfigMapName,
+		Cluster:         cluster,
```
sherine-k, line 118: Great find!

sherine-k (approved):
Great find @ayesha54 for the cluster client! I learned something.
I left some questions and comments here and  there, most important one being : is there a reason we drop the rbac filter by resource name on the configmaps?
but nothing else we cant address in another smaller PR if need be.

pkg/operator/controller/nodeobservability/controller.go
```diff
@@ -41,36 +42,37 @@ import (
...
-	"github.com/openshift/node-observability-operator/pkg/operator/controller/utils"
+	ctrlutils "github.com/openshift/node-observability-operator/pkg/operator/controller/utils"
 )
 
 const (
 	finalizer = "NodeObservability"
 	// the name of the NodeObservability resource which will be reconciled
-	nodeObsCRName = "cluster"
+	nodeObsCRName        = "cluster"
+	defaultRequeuePeriod = time.Duration(5) * time.Second
```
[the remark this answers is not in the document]
ayesha54: We want requeue period separated for each controller.

pkg/operator/controller/utils/test/vars.go
ayesha54: We are using them for tests specifically, I am not really sure if I should use it from names.go

alebedev87:
/label px-approved

As per Chris Fields message, PX is fine if the docs are up to date. For that one there is no docs.

PR #112, opened 2022-10-27
CFE-634: nob controller: reload kubelet CA configmap

alebedev87:
Followup of https://github.com/openshift/node-observability-operator/pull/111.
The ca-configmap controller syncs the kubelet CA configmap from `openshift-config-managed` namespace into NOB operator's namespace. The agent daemonset consumes the synced configmap. However the configmap may change (certificate rotation). The ca-configmap controller would sync the contents and the kubelet (on the node where agent POD is scheduled) would update the contents in the volume mounted by the agent's POD. But the agent's process doesn't know that the CA bundle changed on the filesystem.

The PR aims at bridging this gap. The POD spec is changed every time there are new contents of the kubelet CA configmap from the operator's namespace. The change consist in the update of the annotation with the hash of the configmap's data.


alebedev87:
/label px-approved

As per Chris Fields, the docs are sufficient for this release.

PR #115, opened 2022-11-29
release 0.2.0

alebedev87:
`alpha` channel is kept to force the migration to the new version of the operator.


alebedev87:
/label docs-approved
/label px-approved

No docs impact as the OLM channel didn't change.

PR #116, opened 2022-11-30
No subscription is required for NOB Operator

alebedev87:
CPaaS' CVP test gives a warning about the OCP subscription requirement.


alebedev87:
/label docs-approved
/label px-approved

No docs impact.

PR #128, opened 2025-12-17
Remove unused certs

lmzuccarelli:
Certs in hack folder are outdated and unused


lmzuccarelli:
This will cause our tests to fail. 
The  certs are used for testing purposes only so its ok to have them in the repository

